PR #32, opened 2023-08-21
th-25: 404 Page

### LINKED ISSUE #25: 404 Page
Implement the 404 page. 
The user should see that page if route is not valid or he does not have permissions to view requested page

akymenko-m:
- implemented 404 page


frontend/src/index.tsx
```diff
@@ -33,6 +34,10 @@ createRoot(document.querySelector('#root') as HTMLElement).render(
                 path: AppRoute.SIGN_UP,
                 element: <Auth />,
               },
+              {
```
kalinkaaaa14: let's move RouterProvider out of index, to some separate component

kalinkaaaa14 (changes requested):

frontend/src/index.tsx
akymenko-m: create new separate component Router and add in index.tsx

akymenko-m:
404-page

frontend/src/pages/not-found/not-found.tsx
```diff
@@ -0,0 +1,19 @@
...
+const NotFound = (): JSX.Element => (
+  <div>
+    <h2>
+      Oops, the dreaded <span>404</span>.
+    </h2>
+
+    <p>Let&apos;s get you back on track.</p>
+
+    <p>
+      <NavLink to={AppRoute.ROOT}>Back to Home page</NavLink>
```
kalinkaaaa14: create custom component wrapper Link which will get children, to, and classname through props. This Link component will be used throughout the project

kalinkaaaa14 (changes requested):

frontend/src/pages/not-found/not-found.tsx
akymenko-m: add reusable component Link instead just NavLink. added props className to this component

frontend/src/libs/components/link/link.tsx
```diff
@@ -6,10 +6,17 @@ import { type ValueOf } from '~/libs/types/types.js';
 type Properties = {
   to: ValueOf<typeof AppRoute>;
```
kalinkaaaa14, line 9: why not like this ?
```suggestion
  to: AppRoute;
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -6,10 +6,17 @@ import { type ValueOf } from '~/libs/types/types.js';
...
   children: React.ReactNode;
```
kalinkaaaa14, line 10: you don't need to describe children prop here

frontend/src/libs/components/link/link.tsx
```diff
@@ -6,10 +6,17 @@ import { type ValueOf } from '~/libs/types/types.js';
...
 };
 
-const Link: React.FC<Properties> = ({ children, to }: Properties) => (
-  <NavLink to={to}>{children}</NavLink>
+const Link: React.FC<Properties> = ({
+  children,
+  to,
+  className,
+}: Properties) => (
+  <NavLink to={to} className={className}>
```
kalinkaaaa14: add default styling
```suggestion
  <NavLink to={to} className={className ?? styles.link}>
```

frontend/src/pages/not-found/not-found.tsx
```diff
@@ -0,0 +1,18 @@
+import { Link } from '~/libs/components/components.js';
+import { AppRoute } from '~/libs/enums/app-route.enum';
```
kalinkaaaa14:
```suggestion
import { AppRoute } from '~/libs/enums/enums';
```

kalinkaaaa14 (changes requested):

frontend/src/libs/components/link/link.tsx
akymenko-m: it's was component from starter, and I just supplemented it. I didn't create a new wrapper component, so that there are no two similar ones.

akymenko-m: If indicate AppRoute I get an error, and need to indicate "typeof AppRoute", but than I have errors from all files where use prop to (because it's expected string, not object type). If I specify ValueOf<typeof AppRoute> work correct. So I leave it whiteout changes

frontend/src/libs/components/link/link.tsx
akymenko-m: If remove children prop, I have error in component that there no specify children in props validation. but I need to pass children in prop. So I leave children in type

frontend/src/libs/components/link/link.tsx
akymenko-m: add default styles to NavLink

frontend/src/pages/not-found/not-found.tsx
akymenko-m: thank you! updated

frontend/src/libs/components/link/link.tsx
PavloSukhinin: That's right, it should be `ValueOf<typeof AppRoute>;` because we use const instead of enums

frontend/src/libs/components/router/router.tsx
```diff
@@ -0,0 +1,35 @@
+import { App, RouterProvider } from '~/libs/components/components.js';
```
PavloSukhinin: That's circular dependency. Router imported from components/components.js and has imports from the same file. 
Be careful with that, it can lead to hard to debug bugs.
Use direct import from `../app/app.js`, same for router provider

PavloSukhinin (changes requested):

frontend/src/libs/components/router/router.tsx
akymenko-m: thank you! update imports:
import { App } from '../app/app.js';
import { RouterProvider } from '../router-provider/router-provider.js';

frontend/src/libs/components/link/link.tsx
kalinkaaaa14: oh, sorry, I didn't think about it, thanks!

kalinkaaaa14 (approved):

PavloSukhinin (approved):

PR #33, opened 2023-08-21
th-26: Add toast notifications service

### LINKED ISSUE #26: Toast notifications
Use React-Toastify.

Create the Notifications service which will accumulate all React-Toastify abilities in a single place.

denred:
- Installed the "react-toastify" package to manage notifications.

- Added a new notification service for convenient display of notifications.

- Exported an instance of the notification service into the relevant part of the application.

- Set up default options to configure the standard behavior of notifications.


frontend/src/index.tsx
```diff
@@ -38,5 +40,6 @@ createRoot(document.querySelector('#root') as HTMLElement).render(
         ]}
       />
     </StoreProvider>
+    <ToastContainer />
```
kalinkaaaa14: create a wrapper for toast container and place this wrapper somewhere in components, it should not be placed in index file

frontend/src/libs/packages/notification/libs/options/notification.options.default.ts
```diff
@@ -0,0 +1,16 @@
+import { type ToastOptions, toast } from 'react-toastify';
+
+import { NotificationTheme } from '../enums/notification.theme.enum.js';
+
+const defaultNotificationOptions: ToastOptions = {
```
kalinkaaaa14:
```suggestion
const DEFAULT_NOTIFICATION_OPTIONS: ToastOptions = {
```

frontend/src/libs/packages/notification/notification.ts
```diff
@@ -0,0 +1,8 @@
+import { defaultNotificationOptions } from './libs/options/notification.options.default.js';
+import { NotificationService } from './notification-api.js';
+
+const notification = NotificationService.getInstance(
+  defaultNotificationOptions,
+);
```
kalinkaaaa14: I think we don't need get instance method, use public constructor instead
const notification = new NotificationService(options);

frontend/src/libs/packages/notification/notification-api.ts
```diff
@@ -0,0 +1,43 @@
...
+    toast.info(message, { ...this.#options, ...options });
+  }
+
+  public static getInstance(options: ToastOptions): NotificationService {
+    if (!this.#instance) {
+      this.#instance = new NotificationService(options);
+    }
+
+    return this.#instance;
+  }
```
kalinkaaaa14: think we can get rid of it

frontend/src/libs/packages/notification/libs/options/notification.options.default.ts
kalinkaaaa14: also create separate type for toast options, describe inside only those which we are planning to use. Now a lot of default options I suppose we will not use

frontend/src/libs/packages/notification/libs/enums/notification.theme.enum.ts
```diff
@@ -0,0 +1,9 @@
+import { type Theme } from 'react-toastify';
+
+const NotificationTheme = {
+  LIGHT: 'light' as Theme,
+  DARK: 'dark' as Theme,
+  COLORED: 'colored' as Theme,
+};
```
kalinkaaaa14:
```suggestion
const NotificationTheme: TypeHere = {
  LIGHT: 'light',
  DARK: 'dark',
  COLORED: 'colored',
} as const;
```

frontend/src/libs/packages/notification/notification-api.ts
```diff
@@ -0,0 +1,43 @@
+import { type ToastOptions, toast } from 'react-toastify';
+
+class NotificationService {
+  #options: ToastOptions;
+
+  static #instance: NotificationService | undefined;
```
kalinkaaaa14: can get rid of

kalinkaaaa14 (changes requested):

frontend/src/libs/components/notification/notification.tsx
```diff
@@ -0,0 +1,10 @@
+import 'react-toastify/dist/ReactToastify.css';
```
kalinkaaaa14: do we need this import here?

frontend/src/libs/components/notification/notification.tsx
```diff
@@ -0,0 +1,10 @@
...
+
+import { type ToastContainerProps } from 'react-toastify';
+import { ToastContainer } from 'react-toastify';
```
kalinkaaaa14: should be in 1 import

frontend/src/libs/packages/notification/libs/enums/notification.theme.enum.ts
```diff
@@ -0,0 +1,7 @@
+const theme = {
```
kalinkaaaa14:
```suggestion
const NotificationTheme= {
```
also file naming - notification-theme.enum

kalinkaaaa14 (changes requested):

frontend/src/libs/packages/notification/libs/options/notification.options.default.ts
```diff
@@ -0,0 +1,17 @@
+import { type Theme, type ToastPosition, toast } from 'react-toastify';
+
+import { theme } from '../enums/notification.theme.enum.js';
+
+type IToastOptions = {
+  position?: ToastPosition;
+  theme?: Theme;
```
kalinkaaaa14: you used Theme from the library in type, but in default options - created theme enum. I guess it should be custom enum everywhere

frontend/src/libs/packages/notification/libs/options/notification.options.default.ts
```diff
@@ -0,0 +1,17 @@
...
+  autoClose?: number;
+};
+
+const DEFAULT_NOTIFICATION_OPTIONS: IToastOptions = {
+  position: toast.POSITION.TOP_RIGHT,
+  theme: theme.LIGHT,
+  autoClose: 4000,
```
kalinkaaaa14, line 19: I think we need to add also transition

kalinkaaaa14 (changes requested):

frontend/src/libs/components/notification/notification.tsx
denred: should styles be imported within the component or at the time the component is used, and which approach is better?

frontend/src/libs/packages/notification/libs/options/notification.options.default.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  type ToastPosition,
+  type ToastTransition,
+  toast,
+} from 'react-toastify';
+
+import { NotificationTheme, NotificationTransition } from '../enums/enums.js';
+
+type IToastOptions = {
```
PavloSukhinin, line 9: Let's name types without `I` prefix

frontend/src/libs/packages/notification/libs/options/notification.options.default.ts
```diff
@@ -0,0 +1,23 @@
...
+  position?: ToastPosition;
+  theme?: (typeof NotificationTheme)[keyof typeof NotificationTheme];
```
PavloSukhinin, line 11:
```suggestion
  theme?: ValueOf<typeof NotificationTheme>;
```

PavloSukhinin (changes requested):
Don't forget to pass instance of NotificationService as extra arguments for Thunks
`frontend/src/libs/packages/store/store.package.ts`

PR #40, opened 2023-08-22
th-18: Implement Customer Signup Form

AlexVlaso:
Снимок экрана 2023-08-22 в 13 05 13


frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -12,44 +14,53 @@ type Properties = {
...
-    validationSchema: userSignUpValidationSchema,
-  });
-
-  const handleFormSubmit = useCallback(
-    (event_: React.BaseSyntheticEvent): void => {
-      void handleSubmit(onSubmit)(event_);
-    },
-    [handleSubmit, onSubmit],
-  );
+  const nameField: FormField<UserSignUpRequestDto> = {
```
Druxight: 1) There is no need to name objects, just create an array called `formFields`  
2) Move the field definition out of the component.

Druxight (changes requested):

kalinkaaaa14:
don't forget to re-request reviewers

shared/src/packages/users/libs/types/user-sign-up-request-dto.type.ts
```diff
@@ -1,6 +1,9 @@
 import { type UserEntity } from './user-entity.type.js';
 
 type UserSignUpRequestDto = Pick<UserEntity, 'phone'> & {
```
kalinkaaaa14, line 3: I think you can add name/surname/email to UserEntity

frontend/src/pages/auth/components/sign-up-form/libs/fields.ts
```diff
@@ -0,0 +1,23 @@
+import { type UserSignUpRequestDto } from 'shared/build';
+
+import { type FormField } from '~/libs/types/form.type.js';
+
+const signUpFields: FormField<UserSignUpRequestDto>[] = [
+  { label: 'Name', placeholder: 'Enter your name', name: 'name' },
```
kalinkaaaa14: separate enum for labels, names

shared/src/packages/users/libs/types/user-sign-up-request-dto.type.ts
AlexVlaso: This is a bad idea because with UserEntity we should rewrite the user.service and add migrations. Also as I know another student is working now with the user.service

kalinkaaaa14: got it, thanks!

frontend/src/libs/components/form/form.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { useCallback } from 'react';
+import { type DeepPartial, type FieldValues } from 'react-hook-form';
+import { type ValidationSchema } from 'shared/build/index.js';
```
kalinkaaaa14: this is import from libs/types/types, same as type FormField, unite them

frontend/src/libs/components/form/form.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { useCallback } from 'react';
```
kalinkaaaa14: useCallback should be from ~/libs/hooks/hooks.js'; such things from frameworks/libraries should be imported once directly, then reused.
also import React is not needed

frontend/src/libs/components/form/form.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { type DeepPartial, type FieldValues } from 'react-hook-form';
+import { type ValidationSchema } from 'shared/build/index.js';
+
+import { useAppForm } from '~/libs/hooks/hooks.js';
+import { type FormField } from '~/libs/types/form.type.js';
+
+import { Button } from '../components.js';
```
kalinkaaaa14: it could be circular dependency when you will add Form component to libs/components/components file. 
To avoid circular dep, rewrite this import directly to button, as you are already in components folder --> as you did it with input

frontend/src/libs/components/form/form.tsx
PavloSukhinin: Not resolved

frontend/src/libs/components/form/form.tsx
```diff
@@ -0,0 +1,55 @@
...
+  onSubmit,
+}: Properties<T>): JSX.Element => {
+  const { control, errors, handleSubmit } = useAppForm<T>({
+    defaultValues,
+    validationSchema,
+  });
+
+  const handleFormSubmit = useCallback(
+    (event_: React.BaseSyntheticEvent): void => {
+      void handleSubmit(onSubmit)(event_);
```
PavloSukhinin, line 34: I believe it's not required to additionally eval this call with void. Am I missing something?

frontend/src/pages/auth/components/sign-up-form/libs/fields.ts
```diff
@@ -0,0 +1,32 @@
+import { type UserSignUpRequestDto } from 'shared/build';
```
PavloSukhinin: import from ~/libs/types/types.js

frontend/src/pages/auth/components/sign-up-form/libs/fields.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { FormLabels, FormNames } from '~/libs/enums/form.enum';
+import { type FormField } from '~/libs/types/form.type.js';
```
PavloSukhinin:
```suggestion
import { type FormField } from '~/libs/types/types.js';
```

frontend/src/pages/auth/components/sign-up-form/libs/fields.ts
```diff
@@ -0,0 +1,32 @@
+import { type UserSignUpRequestDto } from 'shared/build';
+
+import { FormLabels, FormNames } from '~/libs/enums/form.enum';
```
PavloSukhinin:
```suggestion
import { FormLabels, FormNames } from '~/libs/enums/enums.js';
```

frontend/src/libs/enums/form.enum.ts
```diff
@@ -0,0 +1,17 @@
+const FormNames = {
+  NAME: 'name',
+  SURNAME: 'surname',
+  EMAIL: 'email',
+  PHONE: 'phone',
+  PASSWORD: 'password',
+} as const;
+
+const FormLabels = {
```
PavloSukhinin: FormName and FormLabel. Name enums using singular nouns

frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -1,55 +1,32 @@
-import { Button, Input } from '~/libs/components/components.js';
-import { useAppForm, useCallback } from '~/libs/hooks/hooks.js';
+import { Link } from '~/libs/components/components.js';
+import { Form } from '~/libs/components/form/form.js';
+import { AppRoute } from '~/libs/enums/app-route.enum.js';
```
PavloSukhinin:
```suggestion
import { Form } from '~/libs/components/components.js';
import { AppRoute } from '~/libs/enums/enums.js';
```

package.json
```diff
@@ -29,6 +29,7 @@
     "@commitlint/config-conventional": "17.4.4",
     "@ls-lint/ls-lint": "2.0.1",
     "@typescript-eslint/eslint-plugin": "5.48.1",
+    "@typescript-eslint/parser": "5.62.0",
```
PavloSukhinin, line 32: You really need it?

PavloSukhinin (changes requested):

frontend/src/libs/components/form/form.tsx
AlexVlaso: This piece of code I took from the starter and without void it gets the error: "Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator."

package.json
AlexVlaso: I don’t know why but without it, lint doesn’t work, and when you try to run "npm run lint:js" it can’t find the parser module. I tried to install the parser globally but lint didn’t see it, on stackoverflow I found  that the last lint versions can’t use plugins globally

frontend/src/libs/components/form/form.tsx
PavloSukhinin: Got it, ok 👍

package.json
PavloSukhinin: Interesting, OK, let's try with installed pkg

PavloSukhinin (approved):

kalinkaaaa14 (approved):

Druxight (approved):

PR #43, opened 2023-08-22
th-21: Create the css selectors for the input components

albsbz:
Added styles for input components


frontend/src/libs/components/input/input.tsx
```diff
@@ -7,13 +7,18 @@ import {
 
 import { useFormController } from '~/libs/hooks/hooks.js';
 
+import styles from './styles.module.scss';
+
 type Properties<T extends FieldValues> = {
   control: Control<T, null>;
   errors: FieldErrors<T>;
   label: string;
```
kalinkaaaa14: there could be cases where label is not needed for input

frontend/src/libs/components/input/input.tsx
```diff
@@ -23,17 +28,65 @@ const Input = <T extends FieldValues>({
...
+    if (isDisabled) {
+      classNames += ' ' + styles.disabled;
+    }
+
+    if (error) {
+      classNames += ' ' + styles.error;
+    }
+
+    return classNames;
+  };
```
kalinkaaaa14: for many classes, we will use clsx, check other prs, someone has added it I saw, so delete this part

frontend/src/libs/components/input/input.tsx
```diff
@@ -23,17 +28,65 @@ const Input = <T extends FieldValues>({
...
+    <label className={styles.label}>
       <span>{label}</span>
-      <input {...field} type={type} placeholder={placeholder} />
-      {hasError && <span>{error as string}</span>}
+      {type !== 'textarea' && (
+        <input
+          {...field}
+          type={type}
+          placeholder={placeholder}
+          className={`${styles.input} ${mapStatuses()}`}
```
kalinkaaaa14: fix using clsx

frontend/src/libs/components/input/input.tsx
```diff
@@ -23,17 +28,65 @@ const Input = <T extends FieldValues>({
...
-    <label>
+    <label className={styles.label}>
       <span>{label}</span>
-      <input {...field} type={type} placeholder={placeholder} />
-      {hasError && <span>{error as string}</span>}
+      {type !== 'textarea' && (
+        <input
+          {...field}
+          type={type}
+          placeholder={placeholder}
```
kalinkaaaa14: create constant inputArea before rendering jsx, and depending which type you get, return needed jsx

frontend/src/libs/components/input/input.tsx
```diff
@@ -23,17 +28,65 @@ const Input = <T extends FieldValues>({
...
+          className={`${styles.input} ${mapStatuses()} ${styles.textarea} `}
+          disabled={isDisabled}
+        />
+      )}
+
+      {type === 'password' && (
+        <span className={styles.passwordEye}>&#128065;</span>
+      )}
+
+      <span className={`${styles.errorMessage} ${hasError ? '' : '.hidden'}`}>
```
kalinkaaaa14: use clsx

frontend/src/libs/components/input/input.tsx
```diff
@@ -23,17 +28,65 @@ const Input = <T extends FieldValues>({
...
+        {error as string}
```
kalinkaaaa14, line 67: which type can be error here besides string? if undefined, so this block should not render at all

kalinkaaaa14 (changes requested):

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,78 @@ import {
...
+      placeholder={placeholder}
+      className={getValidClassNames(...inputStyles, styles.textarea)}
+      disabled={isDisabled}
+    />
+  );
 
   return (
-    <label>
+    <label className={styles.label}>
       <span>{label}</span>
```
kalinkaaaa14: if label is not passed, this element should not be rendered

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,78 @@ import {
...
-      <input {...field} type={type} placeholder={placeholder} />
-      {hasError && <span>{error as string}</span>}
+      {type === 'textarea' ? InputArea : Input}
+      {hasError && (
+        <span className={getValidClassNames(styles.errorMessage)}>
```
kalinkaaaa14: you don't need getValidClassNames when there is only 1 class

frontend/src/libs/components/input/input.tsx
albsbz: I can replace it, with div in such case, because I use it as wrapper for styles?

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,85 @@ import {
...
   const hasError = Boolean(error);
+  const inputStyles = [
+    styles.input,
+    isActive && styles.active,
+    isFilled && styles.filled,
+    isDisabled && styles.disabled,
+    error && styles.error,
+  ];
+  const Input = (
+    <>
```
Druxight: redundant Fragment

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,85 @@ import {
...
+
 type Properties<T extends FieldValues> = {
   control: Control<T, null>;
   errors: FieldErrors<T>;
-  label: string;
+  label?: string;
   name: FieldPath<T>;
   placeholder?: string;
-  type?: 'text' | 'email' | 'password';
+  type?: 'text' | 'email' | 'password' | 'textarea';
```
Druxight: Remove Textarea, there will be another component for that

Druxight (changes requested):

frontend/src/libs/components/input/input.tsx
albsbz: removed

frontend/src/libs/components/input/input.tsx
albsbz: removed, github shows fragment from another place as after editing it at the same line as removed one

frontend/src/libs/components/input/input.tsx
albsbz: removed getValidClassNames for the line

frontend/src/libs/components/input/input.tsx
albsbz: replaced with div when no label

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,71 @@ import {
...
+          <span className={styles.passwordEye}>&#128065;</span>
+        )}
+      </div>
+      {hasError && (
+        <span className={styles.errorMessage}>{error as string}</span>
+      )}
+    </>
+  );
+
+  return label ? (
```
Druxight: What is the purpose that? 

Just return InputContent JSX wrapped by div
Also, there is no need to create InputContent const

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,71 @@ import {
...
+  isDisabled,
 }: Properties<T>): JSX.Element => {
   const { field } = useFormController({ name, control });
 
   const error = errors[name]?.message;
   const hasError = Boolean(error);
+  const inputStyles = [
+    styles.input,
+    isActive && styles.active,
+    isFilled && styles.filled,
```
PavloSukhinin: If I get it right - it's for controlling the focus state, not sure we need isFilled prop at all. Just css `:focus` is enough

Druxight (changes requested):

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,71 @@ import {
...
+  isFilled,
+  isDisabled,
 }: Properties<T>): JSX.Element => {
   const { field } = useFormController({ name, control });
 
   const error = errors[name]?.message;
   const hasError = Boolean(error);
+  const inputStyles = [
+    styles.input,
+    isActive && styles.active,
```
PavloSukhinin: Not sure why we need isActive 🤔

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,71 @@ import {
...
+          className={getValidClassNames(...inputStyles)}
+          disabled={isDisabled}
+        />
+        {type === 'password' && (
+          <span className={styles.passwordEye}>&#128065;</span>
+        )}
+      </div>
+      {hasError && (
+        <span className={styles.errorMessage}>{error as string}</span>
+      )}
```
PavloSukhinin: UI will "jump" after the error block rendered, let's "reserve" some space for the error messages.

PavloSukhinin (changes requested):

frontend/src/libs/components/input/input.tsx
albsbz: Is it should be something like this? [https://github.com/BinaryStudioAcademy/bsa-2023-towhub/pull/43#discussion_r1302134495](url) Adding and removing hidden class but keep element present?

frontend/src/libs/components/input/input.tsx
albsbz: fixed

frontend/src/libs/components/input/input.tsx
albsbz: removed isActive

frontend/src/libs/components/input/input.tsx
albsbz: removed isFilled

frontend/src/libs/components/input/input.tsx
albsbz: Added margin for element before,and absolute position for error element

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,36 +5,62 @@ import {
...
   return (
-    <label>
-      <span>{label}</span>
-      <input {...field} type={type} placeholder={placeholder} />
-      {hasError && <span>{error as string}</span>}
-    </label>
+    <div className={styles.inputComponentWrapper}>
+      {hasLabel && <span className={styles.label}>{label}</span>}
+      <div className={styles.inputWrapper}>
+        <input
```
PavloSukhinin: I believe it still should be wrapped with `<label>` tag (just change one of the div's to label, but don't forget that dev can't be used as child of label)

albsbz: fixed. replaced outter div to label, and inner div to span

frontend/src/libs/components/input/input.tsx
Druxight: There is no need to add absolute position just render empty element

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,61 @@ import {
...
 }: Properties<T>): JSX.Element => {
   const { field } = useFormController({ name, control });
 
   const error = errors[name]?.message;
   const hasError = Boolean(error);
+  const hasValue = Boolean(field.value);
+  const hasLabel = Boolean(label);
+  const inputStyles = [
+    styles.input,
+    hasValue && styles.active,
```
Druxight: Styles active should be applied not when the filed has value, but when field is focused

Druxight (changes requested):

frontend/src/libs/components/input/input.tsx
albsbz: maybe wrong naming for classes, now is filled applied for focus(with blue border) and active just for darker text inside input, but I'll change this

albsbz: fixed

frontend/src/libs/components/input/input.tsx
albsbz: > There is no need to add absolute position just render empty element

fixed

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,67 @@ import {
...
 
   const error = errors[name]?.message;
   const hasError = Boolean(error);
+  const hasValue = Boolean(field.value);
+  const hasLabel = Boolean(label);
+  const inputStyles = [
+    styles.input,
+    hasValue && styles.filled,
+    isDisabled && styles.disabled,
+    error && styles.error,
```
kalinkaaaa14: hasError && styles.error

albsbz: fixed

PavloSukhinin (approved):

kalinkaaaa14 (approved):

PR #44, opened 2023-08-22
th-27: Integrated google maps

MaxHopaytsa:
- Installed the "@googlemaps/react-wrapper" package
- Added a map component 
- Added a service for Google maps with the following features: 
   - distance calculation
   - adding a marker
   - calculation of the direction and time to overcome it 
 
map
**Note:** At this stage, I added the coordinates as mock data for now.I haven't created styles as such yet either, I just added the map size through the prop style.


frontend/src/libs/components/app/app.tsx
```diff
@@ -49,6 +49,9 @@ const App: React.FC = () => {
         <>
           <h2>Users:</h2>
           <h3>Status: {dataStatus}</h3>
+          <div style={{ width: '500px', height: '500px' }}>
```
kalinkaaaa14: don't use inline styles

frontend/src/libs/components/map/map-inner-component/map-inner-component.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useRef } from 'react';
```
kalinkaaaa14: import from '~/libs/hooks/hooks.js';
we need to make import from the library only in 1 place

frontend/src/libs/packages/map/map.packages.ts
kalinkaaaa14, line 1: map.package.ts - file naming

shared/src/libs/packages/map/libs/interfaces/map.interface.ts
```diff
@@ -0,0 +1,13 @@
+interface IMapService {
```
kalinkaaaa14: do you need this interface in shared folder ?

frontend/src/libs/components/app/app.tsx
MaxHopaytsa: I know about it. This div should not be there yet, I added it only for an example of how the map will look like. I can either remove it from there, or I can put it as a style module (but I don't know where to add it yet, because the map will be used on most pages of the application, but it can have different sizes)

shared/src/libs/packages/map/libs/interfaces/map.interface.ts
MaxHopaytsa: No, I don't need it in this case at all, I just followed the project structure (because, for example, I don't think the storage interface is really necessary in the shared folder either). But maybe it will be useful for someone in the future. Should I remove it?

frontend/src/libs/components/app/app.tsx
kalinkaaaa14: got it, thanks

shared/src/libs/packages/map/libs/interfaces/map.interface.ts
kalinkaaaa14: move it to the frontend, because I suppose it will be used only on fe. If we will need it on shared/be - then we will move it. Storage is in shared, because in future it can be reused not only in fe. With maps I think the situation will be different

frontend/src/libs/components/map/map-inner-component/map-inner-component.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useRef } from '~/libs/hooks/hooks.js';
+import { MapService } from '~/libs/packages/map/map.package';
```
Druxight:
```suggestion
import { MapService } from '~/libs/packages/map/map.js';
```

frontend/src/libs/components/map/map.tsx
```diff
@@ -0,0 +1,32 @@
+import { Status, Wrapper } from '@googlemaps/react-wrapper';
+import { type ReactElement, useCallback } from 'react';
+
+import { config } from '~/libs/packages/config/config.js';
+
+import { MapInnerComponent } from './map-inner-component/map-inner-component.js';
+
+const apiMapKey = config.ENV.API.GOOGLE_MAPS_API_KEY;
+const mapProperties = {
```
Druxight: allow to pass properties to the Map component

frontend/src/libs/packages/config/config.package.ts
```diff
@@ -20,6 +20,8 @@ class Config implements IConfig {
       },
       API: {
         ORIGIN_URL: import.meta.env.VITE_APP_API_ORIGIN_URL as string,
+        GOOGLE_MAPS_API_KEY: import.meta.env
+          .VITE_APP_GOOGLE_MAPS_API_KEY as string,
```
Druxight, line 24: Add VITE_APP_GOOGLE_MAPS_API_KEY to the .env.example

frontend/src/libs/hooks/hooks.ts
```diff
@@ -1,6 +1,6 @@
 export { useAppDispatch } from './use-app-dispatch/use-app-dispatch.hook.js';
 export { useAppForm } from './use-app-form/use-app-form.hook.js';
 export { useAppSelector } from './use-app-selector/use-app-selector.hook.js';
-export { useCallback, useEffect, useMemo, useState } from 'react';
+export { useCallback, useEffect, useMemo, useRef, useState } from 'react';
```
Druxight, line 4: import from libs/hooks

Druxight (changes requested):

PR #47, opened 2023-08-22
th-20: Add styled button component

h0wter:
buttons


frontend/src/assets/css/vars.scss
```diff
@@ -9,6 +9,9 @@ $blue-dark: #3563e9;
 $grey-light: #d5dce8;
 $grey: #90a3bf;
 $grey-dark: #7c8798;
+$shadow-light: 0 4px 8px -2px rgb(0 85 255 / 16%);
+$shadow: 0 2px 4px -2px rgb(0 44 132 / 6%);
+$transition: 0.3s ease;
```
Druxight: We do not need a transition here

frontend/src/libs/components/button/button.module.scss
kalinkaaaa14, line 1: name style files styles.module.scss

frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,32 @@
+import { faChevronDown, faPlus } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
kalinkaaaa14: we need to create custom icon wrapper for this library and reuse it here

frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,32 @@
...
 const Button: React.FC<Properties> = ({
   type = 'button',
+  size = 'md',
+  variant = 'contained',
   label,
-}: Properties) => <button type={type}>{label}</button>;
+  disabled = false,
+}: Properties) => (
+  <button
+    className={[styles.btn, styles[size], styles[variant]].join(' ')}
```
kalinkaaaa14: we will use clsx

frontend/src/libs/components/button/button.tsx
h0wter: Did I understand correctly that all icon imports, for example `import { faChevronDown, faPlus } from '@fortawesome/free-solid-svg-icons';`, will be in the wrapper component? To use the wrapper do I need to create an enum for the icon names?

frontend/src/libs/components/button/button.tsx
h0wter: I remember we discussed this. I wanted to change when clsx will be added to the project. Or should I install it myself?

kalinkaaaa14: created a separate task for clsx library. Let's wait for it, I think it will be implemented soon, and then u can reuse it
https://github.com/BinaryStudioAcademy/bsa-2023-towhub/issues/55

h0wter: Got it, thanks.

frontend/src/libs/components/button/button.tsx
kalinkaaaa14: I propose to add Icon component, which will have props like iconName, className for example
and this wrapper will be for svg icons. So you will have a map 'iconNameToSvg' - which will return svg icon by its name and actually this svg icon will return our Icon component.

frontend/src/assets/css/vars.scss
h0wter: Got it.

frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,36 @@
+import { IconName } from '~/libs/enums/icon-name.enum';
```
PavloSukhinin:
```suggestion
import { IconName } from '~/libs/enums/icon-name.enum.js';
```
Please add extension, we'll investigate why linter missing this check

frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,36 @@
...
+
+import { Icon } from '../components.js';
+import styles from './styles.module.scss';
+
 type Properties = {
   label: string;
   type?: 'button' | 'submit';
+  size?: 'sm' | 'md';
+  variant?: 'contained' | 'outlined' | 'text';
+  disabled?: boolean;
```
PavloSukhinin: Please start booleans with be/have verbs (and their forms)

frontend/src/libs/components/icon/icon.mapper.ts
```diff
@@ -0,0 +1,27 @@
...
+  faChevronDown,
+  faChevronLeft,
+  faGear,
+  faLocationDot,
+  faMap,
+  faPlus,
+  faStar,
+} from '@fortawesome/free-solid-svg-icons';
+
+import { IconName } from '~/libs/enums/icon-name.enum';
```
PavloSukhinin: same about extension, add it please

frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,25 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import { type IconName, type IconSize } from '~/libs/enums/enums.js';
+
+import { iconNameToSvg } from './icon.mapper.js';
+
+type Properties = {
+  iconName: (typeof IconName)[keyof typeof IconName];
+  className?: string;
+  size?: (typeof IconSize)[keyof typeof IconSize];
```
PavloSukhinin:
```suggestion
  iconName: ValueOf<typeof IconName>;
  className?: string;
  size?: ValueOf<typeof IconSize>;
```

PavloSukhinin (changes requested):

Druxight (dismissed):

frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,26 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { type ValueOf } from 'shared/build/index.js';
```
PavloSukhinin:
```suggestion
import { type ValueOf } from '~/libs/types/types.js';
```

frontend/src/libs/components/icon/icon.mapper.ts
```diff
@@ -0,0 +1,27 @@
...
+  faGear,
+  faLocationDot,
+  faMap,
+  faPlus,
+  faStar,
+} from '@fortawesome/free-solid-svg-icons';
+
+import { IconName } from '~/libs/enums/icon-name.enum.js';
+
+const iconNameToSvg = {
```
kalinkaaaa14: add type to this map

frontend/src/libs/components/icon/icon.mapper.ts
kalinkaaaa14, line 1: components/icon/maps/icon-name-to-svg.map
components/icon/maps/maps.ts

frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,36 @@
...
+  isDisabled = false,
+  onClick,
+}: Properties) => (
+  <button
+    className={getValidClassNames(styles.btn, styles[size], styles[variant])}
+    type={type}
+    disabled={isDisabled}
+    onClick={onClick}
+  >
+    <Icon iconName={IconName.PLUS} />
```
Druxight: I do not think that we should always render the PLUS and CHEVRON_DOWN icons. 
Let's consider expanding the label configuration to support the next abilities 
1) Optional icon before label 
2) Optional icon after label  

The easiest way is to make a label string or react element, to allow the dev to pass whatever he wants to the button. 
If there are any other ideas you are welcome 

P.S other than that everything seems OK

Druxight (changes requested):

frontend/src/libs/components/button/button.tsx
h0wter: Yes, that's right, it's better to give the ability to pass different icons.

PavloSukhinin (approved):

Druxight (approved):

PR #34, opened 2023-08-22
th-22: Create checkbox toggle radio components

Some14u:
Requires adjustments of colors and fonts. Waiting for corresponding pr.

Three components are done: checkbox, toggle and radio. All three components are derived from a single checkbox (because there is almost no difference in code, the difference is basically cosmetic).

All three can show error message and able to be disabled. I also added a simple animation setup to sass folder and used it to make toggling smoother.

One package was added: clsx to combine css styles
stylelint package was updated to latest in order to pass style linting (there was an issue with outdated linter with that regard).

checkboxes

I forgot to show it on the picture, but all three elements can work without labels if needed.

PR #54, opened 2023-08-22
th-34: Create checkbox toggle radio components

Some14u:
Applied all required changes, found couple bugs and fixed them all.

In order to work properly, this pr changes .stylelintrc.yml, adding couple necessary rules (I've added comments for each change)

This version correctly uses previously merged scss variables from assets.

checkboxes2


kalinkaaaa14:
https://github.com/BinaryStudioAcademy/bsa-2023-towhub/pull/34
why two different prs on the same ticket?
why th-34? ticket is 22, and branch should be also 22. Branch should tell us to which issue number pr is related. Same about pr naming.

kalinkaaaa14:
checkbox should be implemented without checkmark.svg - only styles
also please, split checkbox/switch/radio component - results should be 3 different components

Some14u:
> checkbox should be implemented without checkmark.svg - only styles

Do you mean to move checkmark sign inline?

> also please, split checkbox/switch/radio component - results should be 3 different components

There are 3 different components (checkbox, toggle, radio), they all share the same base. If you want, i can make a "base" component, and derive all three from it.
There all are almost identical in terms of the code, that's why i decided to have a single element as a base. This is to make it easier changing them later, because of the fact that they have similar nature. For example — if you want to enhance error message, you would probably want to do this for all three elements). This plays well with the DRY rule, though.

> why two different prs on the same ticket?
> why th-34? ticket is 22, and branch should be also 22. Branch should tell us to which issue number pr is related. Same about pr naming.

https://bsa2023.slack.com/files/U05H43YTQH0/F05PC7UG449/image.png

Because it was suggested by @PavloSukhinin to make another pr, and the th-22: Create checkbox toggle radio components
is already exists, so supposedly I can't make another one with the same name. Can I?

frontend/src/libs/components/radio/radio.tsx
```diff
@@ -0,0 +1,70 @@
...
+import radio from './radio.module.scss';
+
+const styles = { ...common, ...radio };
+
+type Properties<T extends FieldValues, N extends FieldPath<T>> = {
+  control: Control<T, null>;
+  errors?: FieldErrors<T>;
+  label?: string;
+  name: N;
+  beDisabled?: boolean;
```
PavloSukhinin:
```suggestion
  isDisabled?: boolean;
```
it's one of the`be` verb forms

frontend/src/libs/components/radio/radio.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { useFormController } from '~/libs/hooks/hooks.js';
+
+import common from './common.module.scss';
+import radio from './radio.module.scss';
+
+const styles = { ...common, ...radio };
+
+type Properties<T extends FieldValues, N extends FieldPath<T>> = {
+  control: Control<T, null>;
+  errors?: FieldErrors<T>;
```
PavloSukhinin, line 15: Can we make it required?
```suggestion
  errors: FieldErrors<T>;
```

frontend/src/libs/components/radio/radio.tsx
```diff
@@ -0,0 +1,70 @@
...
+        className={getValidClassNames(styles.label, beDisabled && 'disabled')}
+      >
+        <input
+          className={getValidClassNames(styles.input, styles.radio)}
+          {...field}
+          type="radio"
+          value={value}
+          disabled={beDisabled || undefined}
+          checked={isChecked}
+          aria-disabled={beDisabled || undefined}
```
PavloSukhinin:
```suggestion
          disabled={isDisabled}
          checked={isChecked}
          aria-disabled={isDisabled}
```
I believe that we don't need undefined here, isDisabled has default value

frontend/src/libs/components/radio/radio.tsx
```diff
@@ -0,0 +1,70 @@
...
+        />
+        {hasLabel && <span className={styles.labelText}>{label}</span>}
+      </label>
+      <div
+        className={getValidClassNames(
+          styles.error,
+          !hasError && 'visually-hidden',
+        )}
+        role="alert"
+        aria-hidden={!hasError || undefined}
```
PavloSukhinin:
```suggestion
        aria-hidden={!hasError}
```
Same as above

frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+  name: FieldPath<T>;
+  beDisabled?: boolean;
+};
+
+const Checkbox = <T extends FieldValues>({
+  control,
+  errors,
+  label,
+  name,
+  beDisabled = false,
```
PavloSukhinin:
```suggestion
  isDisabled = false,
```
🤔

frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useFormController } from '~/libs/hooks/hooks.js';
+
+import checkbox from './checkbox.module.scss';
+import common from './common.module.scss';
+
+const styles = { ...common, ...checkbox };
+
+type Properties<T extends FieldValues> = {
+  control: Control<T, null>;
+  errors?: FieldErrors<T>;
```
PavloSukhinin, line 15:
```suggestion
  errors: FieldErrors<T>;
```
Can we make it required?

Druxight (approved):

frontend/src/libs/components/checkbox/common.module.scss
```diff
@@ -0,0 +1,67 @@
+@use "sass:color";
```
Druxight: why do we need that ?

frontend/src/libs/components/checkbox/common.module.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.label:global(.disabled) + .error {
+  color: color.adjust(
+    $red-dark,
+    $lightness: 35%
+  ); // Disabled color for error messages
+}
+
+.label {
+  --color: #{$blue-dark};
```
Druxight: Could you use colors from the vars.scss ?

frontend/src/libs/components/checkbox/common.module.scss
```diff
@@ -0,0 +1,67 @@
...
+  display: flex;
+  flex-direction: column;
+  font-family: $font-family;
+}
+
+.error {
+  color: $red-dark;
+  font-weight: $font-weight-regular;
+  font-size: 12px;
+  line-height: calc(15.12 / 12);
```
Druxight: Do we need calc here ?

frontend/src/libs/components/toggle/toggle.module.scss
```diff
@@ -0,0 +1,33 @@
+@import "src/assets/css/styles.scss";
+
+.toggle {
+  --size: 20px;
+
+  width: calc(var(--size) * 2);
```
Druxight: Why do we need that ?

frontend/src/libs/components/checkbox/checkbox.module.scss
Druxight, line 1: could you use 1 mudule scss file per component ?

Druxight (changes requested):
Accidentally skip the styling files...

frontend/src/libs/components/checkbox/common.module.scss
Some14u: this is for hover color calculation, and the disabled error color calculation
I use a method from this library to calculate lighter shade of the color. This is the main benefit of using sass, otherwise what is the point of it? We could have used --variables instead.

frontend/src/assets/css/animations.scss
```diff
@@ -0,0 +1,11 @@
+$anim-timing-fn-ui: ease-in-out;
```
kalinkaaaa14, line 1: this animation variables will be used somewhere else throughout the project or only in one place? do we need to place them in src/assets?

frontend/src/libs/components/checkbox/checkbox.module.scss
kalinkaaaa14: agree

frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+        className={getValidClassNames(styles.label, beDisabled && 'disabled')}
+      >
+        <input
+          className={getValidClassNames(styles.input, styles.checkbox)}
+          {...field}
+          type="checkbox"
+          defaultChecked={field.value || undefined}
+          disabled={beDisabled || undefined}
+          aria-disabled={beDisabled || undefined}
+          aria-invalid={hasError || undefined}
```
PavloSukhinin:
```suggestion
          defaultChecked={Boolean(field.value)}
          disabled={isDisabled}
          aria-disabled={isDisabled}
          aria-invalid={hasError}
```

frontend/src/libs/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+        />
+        {hasLabel && <span className={styles.labelText}>{label}</span>}
+      </label>
+      <div
+        className={getValidClassNames(
+          styles.error,
+          !hasError && 'visually-hidden',
+        )}
+        role="alert"
+        aria-hidden={!hasError || undefined}
```
PavloSukhinin:
```suggestion
        aria-hidden={!hasError}
```

frontend/src/libs/components/checkbox/checkbox.tsx
kalinkaaaa14:
```suggestion
  isDisabled = false,
```

frontend/src/libs/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+  name: FieldPath<T>;
+  beDisabled?: boolean;
+};
+
+const Toggle = <T extends FieldValues>({
+  control,
+  errors,
+  label,
+  name,
+  beDisabled = false,
```
PavloSukhinin: Same comments as above, please fix this component too

frontend/src/libs/components/checkbox/common.module.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.label:global(.disabled) + .error {
+  color: color.adjust(
+    $red-dark,
+    $lightness: 35%
+  ); // Disabled color for error messages
```
kalinkaaaa14: remove comments

PavloSukhinin (changes requested):

frontend/src/libs/components/checkbox/common.module.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.labelText {
+  color: var(--color-text);
+  transition: $anim-normal color $anim-timing-fn-ui;
+}
+
+.label:hover {
+  --color: #{$blue-light};
+  --color-text: #{color.adjust($black, $lightness: 20%)};
```
kalinkaaaa14: why not simply color + variable from vars file?

frontend/src/libs/components/toggle/toggle.module.scss
```diff
@@ -0,0 +1,33 @@
...
+  min-width: calc(var(--size) * 2);
+  height: var(--size);
+  border: calc(var(--size) / 10) solid var(--color);
+  border-radius: calc(var(--size) / 2);
+}
+
+.toggle::after {
+  --radius: 6px;
```
kalinkaaaa14: you don't need to add static variables in classnames, and then reuse in calc. there is no sense

frontend/src/libs/components/checkbox/common.module.scss
Some14u: @Druxight $blue-dark is taken from it. Or what do you mean?

frontend/src/libs/components/checkbox/common.module.scss
Some14u: of course, It explains what is this value, and how i've obtained it.
изображение

I think line-height: calc(15.12 / 12) is much more self-explanatory than line-height: 1.26, because the next question could be "why 1.26?"

frontend/src/libs/components/checkbox/checkbox.module.scss
Some14u: Yes, I can, should I? The "common" has absolutely identical content, thus having such separation helps to distinguish which parts of checkbox toggle and radio are equal and require equal changes.
This might come in handy, for example if you would ever decide to change something, which is common for all three of them. You simply change one of "common" modules, and then copypaste it to other ones.

frontend/src/libs/components/radio/radio.tsx
Some14u: It was one of your previous requests. I wanted them to be "isDisabled" from the very beginning.

frontend/src/libs/components/radio/radio.tsx
Some14u: @PavloSukhinin yes, but for radio button it makes no sense, because only the last one is required. You can't have a single radio element which is invalid because they all share the same name thus have single source of error.
In other words, either they all have errors, or none of them. But only the last one will display error after itself.

frontend/src/libs/components/radio/radio.tsx
Some14u: without undefined we would have aria-disabled="false" , disabled="false" in our dom. By having undefined I make sure that dom will remove these properties, if beDisabled (isDisabled) is false.

frontend/src/libs/components/checkbox/checkbox.tsx
Some14u: @PavloSukhinin it was your request: https://github.com/BinaryStudioAcademy/bsa-2023-towhub/pull/34#discussion_r1301129907 :) I cried when turning them to "be" as well :)

frontend/src/libs/components/checkbox/checkbox.tsx
Some14u: yes, but maybe you would want a control without error handling?

frontend/src/assets/css/animations.scss
Some14u: If you asking me, i'd answer yes, this will be used by anyone, who would make ui animations.

frontend/src/libs/components/checkbox/common.module.scss
Some14u: because this is more versatile. You can have a single color variable in vars file, while this:
color.adjust($black, $lightness: 20%)}
is just a modifier of the color. There is no point of having like 5 variations of a single color for every need, when we have such powerful tools like color.adjust and others.

Why wouldn't we use them?

frontend/src/libs/components/toggle/toggle.module.scss
Some14u: @kalinkaaaa14 I guess in that case I can add sass variables. Are you ok with that?

frontend/src/libs/components/toggle/toggle.module.scss
Some14u: again, to understand the idea of how the shape sizes and dimensions are related to each other, We can adjust --size and it will automatically recalculate the shape of the toggle.
For example by looking the formula of width and height you could easily say that width is twice as large as height, and looking down the code you would see how the toggle circle is calculated based on the size.

But maybe you would agree me to use sass variables instead?

frontend/src/libs/components/radio/radio.tsx
Some14u: This is how you suggest:
изображение

This is how is done right now:
изображение

Druxight (dismissed):

package-lock.json
```diff
@@ -2573,7 +2573,7 @@
       "version": "18.2.0",
       "resolved": "https://registry.npmjs.org/@types/react-dom/-/react-dom-18.2.0.tgz",
       "integrity": "sha512-8yQrvS6sMpSwIovhPOwfyNf2Wz6v/B62LFSVYQ85+Rq3tLsBIG7rP5geMxaijTUxSkrO6RzN/IRuIAADYQsleA==",
-      "dev": true,
+      "devOptional": true,
```
Druxight: rollback it

Druxight (approved):

PavloSukhinin (approved):

Some14u:
Hurray!
Sorry, can't hold myself anymore:)

PR #57, opened 2023-08-23
th-48: Dropdown component

DmytroVoronkov:
- Created reusable Dropdown element
image


kalinkaaaa14:
let's add react-select library and customize it
also in the future we will need multiselect, so add such a possibility

frontend/src/libs/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+    { value: 'chocolate', label: 'Chocolate' },
+    { value: 'strawberry', label: 'Strawberry' },
+    { value: 'vanilla', label: 'Vanilla' },
+  ];
+
+  return (
+    // <div className={styles.dropdown}>
+    //   <div className={`${styles['select-wrapper']}`}>
+    //     <select className={styles.select} name={name} id={id}>
+    //       {options.map((option) => (
```
Druxight, line 25: Remove comments
Style the select

Druxight (changes requested):

PR #58, opened 2023-08-23
th-55: Add clsx package for applying multiple classes

denred:
- Add helper to apply multiple classes using the clsx library

### DIFF
--- a/frontend/src/libs/helpers/classname-helper.ts
+++ b/frontend/src/libs/helpers/classname-helper.ts
@@ -1,7 +1,0 @@
-import clsx from 'clsx';
-
-const getClassNames = (...classNames: string[]): string => {
-  return clsx(...classNames);
-};
-
-export { getClassNames };
--- a/frontend/src/libs/helpers/classname-helper.ts
+++ b/frontend/src/libs/helpers/classname-helper.ts
@@ -1,0 +1,7 @@
+import clsx from 'clsx';
+
+const getClassNames = (...classNames: string[]): string => {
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -49,5 +49,8 @@
   "simple-git-hooks": {
     "pre-commit": "npx lint-staged",
     "commit-msg": "npx commitlint --edit $1"
+  },
+  "dependencies": {
+    "clsx": "2.0.0"
```
kalinkaaaa14: should be on fe package.json

frontend/src/libs/helpers/classname-helper.ts
kalinkaaaa14, line 1: should be separate folder get-valid-classnames/get-valid-classnames.helper -- view example of the structure in shared

frontend/src/libs/helpers/classname-helper.ts
```diff
@@ -0,0 +1,7 @@
+import clsx from 'clsx';
+
+const getClassNames = (...classNames: string[]): string => {
```
kalinkaaaa14: add 
```suggestion
const getValidClassNames = (...classNames: ClassValue[]): string => {
```

kalinkaaaa14 (changes requested):

PavloSukhinin (approved):

Druxight (approved):

kalinkaaaa14 (changes requested):
package.json packages should not be changed manually, use npm

PR #60, opened 2023-08-23
th-24: Add protected routing [FE]

h0wter:
I used Math.random as an imitation of the user's login and refresh state. When the redux store is completed, I will update the component.

Created a pull request to make sure I'm moving in the right direction.


frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,22 @@
+import { Navigate } from 'react-router-dom';
+
+import { AppRoute } from '~/libs/enums/enums.js';
+
+type Properties = {
+  component: React.FC;
+};
+
+const ProtectedRoute: React.FC<Properties> = ({
```
Druxight: Please use the `Outlet`  and create the example route called '/protected'

Druxight (changes requested):

frontend/src/libs/components/protected-route/protected-route.tsx
h0wter: Of course, thank you.

h0wter:
Changed the routes declaration, used JSX approach.

In the `ProtectedRoute` component, I used `Navigate`, should it be moved to `components.ts` and exported from there, similar to `RouterOutlet`?

frontend/src/libs/components/router/router.tsx
```diff
@@ -1,34 +1,31 @@
...
+            <Route path={AppRoute.ROOT} element={<App />}>
+              <Route path={AppRoute.SIGN_IN} element={<Auth />} />
+              <Route path={AppRoute.SIGN_UP} element={<Auth />} />
+            </Route>
+            <Route path={AppRoute.ROOT} element={<ProtectedRoute />}>
+              <Route path="protected1" element={<div>Protected path #1</div>} />
+              <Route path="protected2" element={<div>Protected path #2</div>} />
+            </Route>
+            <Route path={AppRoute.ANY} element={<NotFound />} />
+          </Routes>
```
h0wter: If ProtectedRoute needs to be nested in the App layout, I will move it there without any problems.
```suggestion
          <Routes>
            <Route path={AppRoute.ROOT} element={<App />}>
              <Route path={AppRoute.SIGN_IN} element={<Auth />} />
              <Route path={AppRoute.SIGN_UP} element={<Auth />} />
              <Route element={<ProtectedRoute />}>
                <Route
                  path="protected1"
                  element={<div>Protected path #1</div>}
                />
                <Route
                  path="protected2"
                  element={<div>Protected path #2</div>}
                />
              </Route>
            </Route>
            <Route path={AppRoute.ANY} element={<NotFound />} />
          </Routes>
```

frontend/src/libs/components/router/router.tsx
```diff
@@ -1,34 +1,31 @@
...
-          },
-          {
-            path: AppRoute.ANY,
-            element: <NotFound />,
-          },
-        ],
+        path: '*',
+        element: (
+          <Routes>
+            <Route path={AppRoute.ROOT} element={<App />}>
```
Druxight: we do not need this wrapper

frontend/src/libs/components/router/router.tsx
```diff
@@ -1,34 +1,31 @@
...
 import { Auth } from '~/pages/auth/auth.js';
 import { NotFound } from '~/pages/not-found/not-found.js';
 
 import { App } from '../app/app.js';
+import { ProtectedRoute } from '../components.js';
 import { RouterProvider } from '../router-provider/router-provider.js';
 
 const Router = (): JSX.Element => (
   <RouterProvider
     routes={[
```
Druxight: There is no need to use routes={} since you have added <Routes>
```
<RouterProvider>
<Routes>
...
</Routes>
<RouterProvider/>
```

Druxight (changes requested):

frontend/src/libs/components/router/router.tsx
h0wter: Done.

frontend/src/libs/components/router/router.tsx
h0wter: I found only one way to get rid of the wrapper, via `createRoutesFromElements`, but for that I had to remove `<Routes></Routes>` so that this method could build the routes itself.

Druxight (dismissed):

frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,22 @@
+import { Navigate } from 'react-router-dom';
+
+import { AppRoute } from '~/libs/enums/enums.js';
+
+import { RouterOutlet, Spinner } from '../components.js';
+
+const ProtectedRoute = (): React.ReactElement | null => {
+  const isLoggedIn = Math.random();
```
kalinkaaaa14: I think you need to make changes in store, and retrieve user from the store. I know that now there's  no log in functionality, but I don't like the idea to merge this approach with Math.random into development 🤔

h0wter: I did this to test how it works and created a pull request so you can see the rest of the code. It was not supposed to be added in this form to development.

If this component is needed now, then I will do as you say and update it.

kalinkaaaa14: I think you need to add changes in this pr, @PavloSukhinin @Druxight what are u thinking about this?

PavloSukhinin: Sure, we're waiting for a jwt and BE sign-in route implementation and then will ask @h0wter to adapt ProtectedRoute

frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,32 @@
+import { Navigate } from 'react-router-dom';
+import { type UserGroupKey } from 'shared/build/index.js';
```
PavloSukhinin: from '~/packages/users/enums/enums.js'

frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { AppRoute, DataStatus } from '~/libs/enums/enums.js';
+import { useAppSelector } from '~/libs/hooks/hooks.js';
+import { type ValueOf } from '~/libs/types/types.js';
+import { selectIsLoading, selectUser } from '~/slices/auth/selectors.js';
+
+import { RouterOutlet, Spinner } from '../components.js';
```
PavloSukhinin: That's circular dependency. You can re-write it as:
from '../spinner/spinner.js'
from '../router/router.js' // yeah, let's import RouterOutlet and Navigate there.

shared/src/packages/users/libs/types/user.type.ts
```diff
@@ -0,0 +1,15 @@
+import { type BusinessEntityT } from '../../../../packages/business/business.js';
```
PavloSukhinin:
```suggestion
import { type BusinessEntityT } from '~/packages/business/business.js';
```

frontend/src/slices/auth/selectors.ts
```diff
@@ -0,0 +1,11 @@
+import { type User } from 'shared/build/index.js';
+
+import { type DataStatus } from '~/libs/enums/data-status.enum.js';
+import { type RootState, type ValueOf } from '~/libs/types/types.js';
+
+const selectIsLoading = (state: RootState): ValueOf<typeof DataStatus> =>
+  state.auth.dataStatus;
```
PavloSukhinin:
```suggestion
const selectIsLoading = (state: RootState): ValueOf<typeof DataStatus> => {
 return  state.auth.dataStatus;
 }
```
to avoid such line breaks

frontend/src/slices/auth/selectors.ts
```diff
@@ -0,0 +1,11 @@
+import { type User } from 'shared/build/index.js';
+
+import { type DataStatus } from '~/libs/enums/data-status.enum.js';
```
PavloSukhinin:
```suggestion
import { type DataStatus } from '~/libs/enums/enums.js';
```

frontend/src/slices/auth/selectors.ts
```diff
@@ -0,0 +1,11 @@
+import { type User } from 'shared/build/index.js';
```
PavloSukhinin:
```suggestion
import { type User } from '~/packages/users/users.js';
```

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -1,4 +1,5 @@
 import { createSlice } from '@reduxjs/toolkit';
+import { type User } from 'shared/build/index.js';
```
PavloSukhinin:
```suggestion
import { type User } from '~/packages/users/users.js';
```

frontend/src/libs/components/router/router.tsx
```diff
@@ -1,52 +1,39 @@
+import { Route } from 'react-router-dom';
+import { UserGroupKey } from 'shared/build/index.js';
```
PavloSukhinin: from '~/packages/users/enums/enums.js'

PavloSukhinin (changes requested):

frontend/src/libs/components/protected-route/protected-route.tsx
h0wter: Changed it. Did I understand correctly that if a component is exported from components.ts, then inside this component I should not import from components.ts?

PavloSukhinin: Yes, that's correct 👍

shared/src/packages/users/libs/types/user-sign-in-response-dto.type.ts
```diff
@@ -1,9 +1,11 @@
+import { type BusinessEntityT } from '~/packages/business/business.js';
+
 import { type UserEntityT } from './user-entity.type.js';
 import { type UserGroupEntityT } from './user-group-entity.type.js';
 
 type UserSignInResponseDto = Omit<
   UserEntityT,
   'passwordHash' | 'passwordSalt' | 'groupId'
-> & { group: UserGroupEntityT };
+> & { group: UserGroupEntityT } & { business: BusinessEntityT[] };
```
PavloSukhinin:
```suggestion
> & { group: UserGroupEntityT; business: BusinessEntityT[] };
```

PavloSukhinin (dismissed):

PavloSukhinin (approved):

PR #69, opened 2023-08-23
th-27: Integrate Google Maps

MaxHopaytsa:
- Installed @googlemaps/react-wrapper package for easier implementation of Google maps
- Added Google maps component
- Added the MapService class to implement the main functions of the map


.eslintrc.yml
```diff
@@ -125,13 +125,13 @@ rules:
 
 overrides:
   - files:
-    - commitlint.config.ts
+      - commitlint.config.ts
```
Druxight: please check you editor settings and rollback this changes

.github/labeler.yml
```diff
@@ -6,13 +6,13 @@ shared: 'shared/**/*'
 
 backend:
   - any:
-    - 'backend/**/*'
-    - '!backend/.env.example'
+      - 'backend/**/*'
```
Druxight: same here

frontend/.eslintrc.yml
```diff
@@ -23,12 +23,12 @@ rules:
 
 overrides:
   - files:
-    - vite.config.ts
+      - vite.config.ts
```
Druxight: same here

frontend/src/libs/components/map/map.tsx
```diff
@@ -0,0 +1,36 @@
...
+type Properties = MapType & {
+  className?: string;
+};
+
+const Map: React.FC<Properties> = ({ ...mapProperties }) => {
+  const renderMap = useCallback(
+    (status: Status): React.ReactElement => {
+      switch (status) {
+        case Status.LOADING: {
+          return <div>Loading...</div>;
```
Druxight: we have a loading component for that

frontend/src/libs/types/map.type.ts
```diff
@@ -0,0 +1,8 @@
+type MapType = {
```
Druxight: Not sure that we need that. So move all of that to the Map Component 
Also, it may cause application errors because there is no  google.maps until the google script is loaded

Druxight (changes requested):

frontend/src/libs/components/map/map.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import { useCallback } from '~/libs/hooks/hooks.js';
+import { config } from '~/libs/packages/config/config.js';
+
+import { Spinner } from '../components.js';
+import { MapInnerComponent } from './map-inner-component/map-inner-component.js';
+
+const apiMapKey = config.ENV.API.GOOGLE_MAPS_API_KEY;
+
+type MapType = {
```
Druxight: Since you are not using it anywhere except component props lets remove it and move everything to the properties

Druxight (changes requested):

Druxight (approved):

PavloSukhinin (approved):

PR #71, opened 2023-08-23
th-14: Create header component

Some14u:
A simplest ever possible header. 
изображение

The logo consists of two svg images. Logo component accepts optional prop called "layout", you can pass "horizonal" or "vertical". Vertical layout show the brand name under the logo symbol.

Also i made logo in that way so as long as the container has specified width, it will adjust it's size respectively. This way you can make logo of any size by explicitly specifying the width of it's container:
изображение


frontend/src/libs/components/app-logo/app-logo.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type AppLogoProperties = {
+  layout?: 'horizontal' | 'vertical';
+};
+
+const AppLogo: React.FC<AppLogoProperties> = ({
+  layout = 'horizontal',
+}: AppLogoProperties) => {
+  return (
+    <div className={[styles.container, styles[layout]].join(' ')}>
```
Druxight: use clsx for that

frontend/src/libs/components/app/app.tsx
```diff
@@ -27,9 +27,15 @@ const App: React.FC = () => {
     }
   }, [isRoot, dispatch]);
 
+  const handleSighIn = useCallback(() => {
```
Druxight: handleSignIn function should be inside Header component

frontend/src/libs/components/app/app.tsx
```diff
@@ -27,9 +27,15 @@ const App: React.FC = () => {
...
+    return;
+  }, []);
+
   return (
     <>
-      <img src={reactLogo} className="App-logo" width="30" alt="logo" />
+      <header>
```
Druxight: The header element should be in the header component

Druxight (changes requested):

frontend/src/libs/components/app/app.tsx
Some14u: Great, 100% agree. It was outside in our thread.js task though.

frontend/src/libs/components/app-logo/app-logo.tsx
Some14u: did it

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,30 @@
+import { NavLink } from 'react-router-dom';
```
Druxight: We have a custom component Link for that (frontend/src/libs/components/link/link.tsx)

Druxight (changes requested):

frontend/src/libs/components/header/header.tsx
Some14u: i see, let me use it then

Druxight (approved):

PavloSukhinin (approved):
Button should be replaced later

PR #74, opened 2023-08-24
th-48: Dropdown component

DmytroVoronkov:
Add reusable dropdown component
image
image


PavloSukhinin (commented):
Let's improve styling - component shouldn't have fixed width, add some margin for options

Druxight:
In my view we should get rid of focus border

frontend/src/libs/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const Dropdown: React.FC<Properties> = ({
+  options,
+}: Properties): JSX.Element => {
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+
+  const onMenuOpen = useCallback(() => setIsMenuOpen(true), []);
+  const onMenuClose = useCallback(() => setIsMenuOpen(false), []);
+
+  const stylesConfig: StylesConfig = {
```
Druxight: Move it out of FC and create function `getStyles({isMenuOpen}): StylesConfig` above the component.
Optimize function calls by using useCallback

Druxight (changes requested):

frontend/src/libs/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }),
+  };
+};
+
+const Dropdown: React.FC<Properties> = ({
+  options,
+}: Properties): JSX.Element => {
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+
+  const toggleMenu = useCallback(
```
PavloSukhinin: `const handleToggleMenu`declaration of handlers/callbacks should start with `handle` word. 
callbacks passed via props should start with `on` (just like in Select below -   onMenuOpen)

PavloSukhinin (changes requested):

Druxight (dismissed):

frontend/src/libs/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,73 @@
+import Select, { type StylesConfig } from 'react-select';
+
+import { useCallback, useMemo, useState } from '~/libs/hooks/hooks.js';
+
+type Option = { label: string; value: string };
```
PavloSukhinin: Let's move it to `~/libs/types` and rename to SelectOption. We'll need it in case we want to declare options in page domain.

PavloSukhinin (changes requested):

frontend/src/libs/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,76 @@
+import Select, { type StylesConfig } from 'react-select';
+
+import { useCallback, useMemo, useState } from '~/libs/hooks/hooks.js';
+import { type SelectOption } from '~/libs/types/select-option.type';
```
PavloSukhinin: file extension, please

PavloSukhinin (approved):

Druxight (approved):

PR #83, opened 2023-08-25
th-9: Rewrite ER model using Mermaid in readme

albsbz:
Rewrite ER model using Mermaid in readme, including implemented tables 

er


readme.md
```diff
@@ -40,7 +40,130 @@ TBA
 
 ### 💽 DB Schema
 
-![db schema](https://github-production-user-asset-6210df.s3.amazonaws.com/19575839/261530664-6140d0e8-abf8-4fb8-80c9-ff9eb7adfcb6.png)
+```mermaid
+
+erDiagram
```
Druxight, line 47: please remove tables that were not implemented yet

Druxight (changes requested):

readme.md
albsbz: done

readme.md
```diff
@@ -42,6 +42,45 @@ TBA
 
 ![db schema](https://github-production-user-asset-6210df.s3.amazonaws.com/19575839/261530664-6140d0e8-abf8-4fb8-80c9-ff9eb7adfcb6.png)
 
+```mermaid
+
+erDiagram
+
+    USERS {
```
PavloSukhinin: Let's name tables in a lower case, just like in the db?

PavloSukhinin (changes requested):

readme.md
albsbz: Done

PavloSukhinin (approved):

readme.md
```diff
@@ -42,6 +42,45 @@ TBA
...
 ![db schema](https://github-production-user-asset-6210df.s3.amazonaws.com/19575839/261530664-6140d0e8-abf8-4fb8-80c9-ff9eb7adfcb6.png)
 
+```mermaid
+
+erDiagram
+
+    users {
+        id serial PK "not null"
+        phone varchar "not null, unique"
+        email varchar "not null"
```
kalinkaaaa14, line 52: Email is also unique

readme.md
```diff
@@ -42,6 +42,45 @@ TBA
...
+    business_details {
+        id serial PK "not null"
+        company_name varchar "not null, unique"
+        tax_number varchar "not null, unique"
+        owner_id integer FK "not null"
+        created_at timestamp "not null"
+        updated_at timestamp "not null"
+    }
+
+    users one or many -- one groups: "users(group_id) belongs to groups(id)"
```
kalinkaaaa14, line 79: I believe something is wrong with relations. Now in your er it's described that 1 user can have multiple businesses, but our app will not work that way 🤔
We discussed two variants:
1 - one user can have one business account, and one business account belongs to 1 user, in that case we can place fk to business details table ( I believe we started with implementing this logic 🤔)
2 - one user can have one business account, and one business account can have multiple users, in this case fk business_id should be placed in user table

kalinkaaaa14 (changes requested):

readme.md
albsbz: Yes, but this is what implemented at the moment, we have set foreign keys between business_details and user tables but not relations, and now nothing stops several business_details have same userId. From the start I added all tables and relations as in start schema, but than it was mentioned that only implemented logic should be present, so when tables relations will be added this PR can be updated or new one created.

readme.md
albsbz: This is not in development branch yet

PavloSukhinin (approved):

readme.md
kalinkaaaa14: Ok, got it

kalinkaaaa14 (approved):

PR #84, opened 2023-08-25
th-65: Create a truck schema

denred:
- Add Trucks Table Schema
- Add UsersTrucks Schema
- Generate Migrations
- Add Truck Repository


backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -39,4 +39,36 @@ const groups = pgTable('groups', {
...
+    return {
+      uniqueLicensePlateNumber: uniqueIndex(
+        'trucks_license_plate_number_idx',
+      ).on(trucks.licensePlateNumber),
+    };
+  },
+);
+
+const usersTrucks = pgTable('users_trucks', {
+  id: serial('id').primaryKey(),
```
albsbz: Maybe we can use [Composite Primary Key](https://orm.drizzle.team/docs/indexes-constraints#composite-primary-key) here?

backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -39,4 +40,41 @@ const groups = pgTable('groups', {
...
-export { groups, users };
+const trucks = pgTable(
+  'trucks',
+  {
+    id: serial('id').primaryKey(),
+    createdAt: timestamp('created_at').notNull().defaultNow(),
+    updatedAt: timestamp('updated_at').notNull().defaultNow(),
+    manufacturer: varchar('manufacturer').notNull(),
+    capacity: varchar('capacity').notNull(),
+    pricePerKm: varchar('price_per_km').notNull(),
```
albsbz: if capacity will be in kg, may be better it to make integer, and for price too

backend/src/packages/trucks/truck.entity.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  private createdAt: Date;
+
+  private updatedAt: Date;
+
+  private manufacturer: string;
+
+  private capacity: string;
+
+  private pricePerKm: string;
```
albsbz: same about number

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public async find(id: number): Promise<TruckEntity | null> {
+    const result: TruckEntityT[] = await this.db
+      .driver()
+      .select()
+      .from(this.trucksSchema)
+      .where(eq(this.trucksSchema.id, id));
+
+    return result.length === 0 ? null : TruckEntity.initialize(result[0]);
```
PavloSukhinin:
```suggestion
    const [truck = null]: TruckEntityT[] = await this.db
      .driver()
      .select()
      .from(this.trucksSchema)
      .where(eq(this.trucksSchema.id, id));

    return truck ? TruckEntity.initialize(truck) : null;
```
Don't need to fix this now, because we decided to change approach. Just FYI

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public async findAll(): Promise<TruckEntity[]> {
+    try {
+      const result = await this.db.driver().select().from(this.trucksSchema);
+
+      return result.length === 0
+        ? []
+        : result.map((it) => TruckEntity.initialize(it));
+    } catch (error) {
```
PavloSukhinin:
```suggestion
    } catch (error: unknown) {
```

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+      throw new Error(`Failed to retrieve trucks: ${(error as Error).message}`);
```
PavloSukhinin: - when this error can occur? 
- create custom exception for it

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+            ilike(this.trucksSchema.capacity, `%${query}%`),
+            ilike(this.trucksSchema.pricePerKm, `%${query}%`),
+            ilike(this.trucksSchema.licensePlateNumber, `%${query}%`),
+            ilike(this.trucksSchema.year, `%${query}%`),
+          ),
+        );
+
+      return result.length === 0
+        ? []
+        : result.map((it) => TruckEntity.initialize(it));
```
PavloSukhinin: Redundant ternary - `[].map(... )` will return `[]`

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+          pricePerKm,
+          licensePlateNumber,
+          year,
+        })
+        .returning()
+        .execute();
+
+      return TruckEntity.initialize(result);
+    } catch (error) {
+      throw new Error(
```
PavloSukhinin: same question as above - when this can occur and custom error if needed

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+        .delete(this.trucksSchema)
+        .where(eq(this.trucksSchema.id, id))
+        .returning()
+        .execute();
+
+      return Boolean(item);
+    } catch (error) {
+      throw new Error(
+        `Failed to update truck with ID ${id}: ${(error as Error).message}`,
+      );
```
PavloSukhinin: FYI, same as above

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+
+      return result.length === 0
+        ? []
+        : result.map((it) => TruckEntity.initialize(it));
+    } catch (error) {
+      throw new Error(
+        `An error occurred while searching ${query} for trucks: ${
+          (error as Error).message
+        }`,
+      );
```
PavloSukhinin: FYI, same as above

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,170 @@
...
+        .orderBy(this.trucksSchema.id)
+        .execute();
+
+      return result.map((it) => TruckEntity.initialize(it.trucks));
+    } catch (error) {
+      throw new Error(
+        `Failed to fetch trucks by owner (ownerId: ${ownerId}). Error: ${
+          (error as Error).message
+        }`,
+      );
```
PavloSukhinin: FYI

shared/src/packages/trucks/libs/types/truck-entity.type.ts
```diff
@@ -0,0 +1,12 @@
+type TruckEntity = {
+  id: number;
+  createdAt: Date;
+  updatedAt: Date;
```
PavloSukhinin: for backend when retrieved from DB, it will be Date, but though network it will be passed as string. So on frontend it will be string

PavloSukhinin (changes requested):

backend/src/packages/trucks/libs/types/truck-entity-database-type/truck-entity-database.type.ts
```diff
@@ -0,0 +1,8 @@
+import { type TruckEntity as TruckEntityT } from '../types.js';
```
PavloSukhinin: This is circular dependency, be careful with such imports

backend/src/packages/trucks/truck.entity.ts
```diff
@@ -0,0 +1,111 @@
...
+    manufacturer,
+    capacity,
+    pricePerKm,
+    licensePlateNumber,
+    year,
+  }: Omit<TruckEntityT, 'id' | 'createdAt' | 'updatedAt'>): TruckEntity {
+    return new TruckEntity({
+      id: null,
+      createdAt: new Date(),
+      updatedAt: new Date(),
```
PavloSukhinin: Let's omit these 2. I believe these should be database tech properties, let's not use them for now in Entities

backend/src/packages/trucks/truck.entity.ts
```diff
@@ -0,0 +1,111 @@
...
+      licensePlateNumber,
+      year,
+    });
+  }
+
+  public toObject(): TruckEntityT {
+    return {
+      id: this.id as number,
+      createdAt: this.createdAt.toISOString(),
+      updatedAt: this.updatedAt.toISOString(),
```
PavloSukhinin: Same, let's remove dates from here

backend/src/packages/trucks/truck.repository.ts
```diff
@@ -0,0 +1,162 @@
...
+        .driver()
+        .select()
+        .from(this.trucksSchema)
+        .leftJoin(
+          this.usersTrucksSchema,
+          eq(this.trucksSchema.id, this.usersTrucksSchema.truckId),
+        )
+        .where(eq(this.usersTrucksSchema.userId, ownerId))
+        .orderBy(this.trucksSchema.id)
+        .execute();
```
PavloSukhinin: That's OK, but maybe we should try with relational queries?

shared/src/packages/trucks/libs/types/truck-entity.type.ts
```diff
@@ -0,0 +1,12 @@
+type TruckEntity = {
+  id: number;
+  createdAt: string;
+  updatedAt: string;
```
PavloSukhinin: Let's remove dates for now?

PavloSukhinin (changes requested):

PR #93, opened 2023-08-25
th-20: Extends a button component

h0wter:
Removed the fixed height of the button, now it is regulated by paddings.
Added ability to set button width: 100% via props.
Also added the ability to pass additional classes to the component.

If we need the ability to change the background of a button via a passed class, we need to use `:where`. I have the code ready and I can update the component if necessary.


frontend/src/libs/components/button/button.tsx
```diff
@@ -6,36 +6,48 @@ import { Icon } from '../components.js';
...
-  </button>
-);
+}: Properties) => {
+  return (
+    <button
+      className={getValidClassNames(
+        styles.btn,
+        styles[size],
+        styles[variant],
+        isFullWidth ? styles.fullWidth : null,
```
PavloSukhinin: isFullWidth && styles.fullWidth

PavloSukhinin (dismissed):

Druxight (approved):

PavloSukhinin (approved):

PR #94, opened 2023-08-25
th-86: Modal component

DmytroVoronkov:
- Created Portal component
- Created Modal component
image


frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
DmytroVoronkov: I did not want to create unneccessary callbacks, so i added this line

frontend/src/libs/components/portal/portal.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import styles from './styles.module.scss';
+
+type Properties = {
+  children: ReactNode;
+};
+
+const Portal: React.FC<Properties> = ({ children }: Properties) => {
+  const portalContainer = useMemo(() => {
+    const element = document.createElement('div');
```
h0wter: Why not add an element for the portal directly to the `index.html`?
For example:
```
   <body>
     <div id="root"></div>
     <div id="modal-root"></div>
     <script type="module" src="/src/index.tsx"></script>
   </body>
```

h0wter:
Maybe add the ability to close the modal with `Esc` button?
Something like:
```
  const onKeyDown = useCallback(
    (e: KeyboardEvent) => {
      if (e.code === "Escape") {
        onClose();
      }
    },
    [onClose]
  );

  useEffect(() => {
    window.addEventListener("keydown", onKeyDown);

    return () => window.removeEventListener("keydown", onKeyDown);
  }, [onKeyDown]);
```

frontend/src/libs/components/modal/styles.module.scss
```diff
@@ -0,0 +1,56 @@
+.modal {
```
PavloSukhinin, line 1: Note for wide screens: I believe we should make modal width same as component inside it. Otherwise there will be a lot of empty space.

PavloSukhinin (changes requested):

frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <dialog
+        className={styles.content}
+        onClick={handleDisableContentContainerClick}
+        onKeyDown={handleExitKeydown}
+        role="button"
+        tabIndex={0}
+      >
+        {children}
+      </dialog>
+    </div>
```
PavloSukhinin: I thought we'll use `<dialog>` for modal itself, not just content

PavloSukhinin (changes requested):

PavloSukhinin (approved):

PR #103, opened 2023-08-26
th-82: Create badge component

albsbz:
Badges
Added badge component that accepts color from pallete and use contrast text color. To use SCSS loops and import scss variables in React component added some changes to stylelint.


frontend/src/.stylelintrc.yml
```diff
@@ -6,18 +6,25 @@ plugins:
...
+    - true
+    - ignoreSelectors:
+        - :export
   # allow @use to precede @import
   no-invalid-position-at-import-rule:
     - true
     - ignoreAtRules:
         - use
   # allow :global which is specific pseudo class, available in .module.css
+  # allow :export which is specific pseudo class, available in .module.css
```
Some14u: export isn't specific pseudo class for .module.css
It was my comment about ":global" pseudo, which is actually is part of .module.css logic and by default is unknown to css, scss and stylelint.

This is only comment, so you don't really have to change anything, but if you would have a chance, please, fix it :)

albsbz: fixed

frontend/src/libs/components/badge/badge.tsx
```diff
@@ -0,0 +1,49 @@
...
+  className?: string;
+  color?: Color;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+};
+
+const Badge: React.FC<Properties> = ({
+  children,
+  className,
+  color = defaultColor,
+  onClick: handleClick,
```
PavloSukhinin: Not necessary to rename handler here. We should name it with `handle` prefix only during declaration.

package-lock.json
```diff
@@ -10511,6 +10496,21 @@
         "stylelint": "^14.0.0 || ^15.0.0"
       }
     },
+    "node_modules/stylelint-scss": {
```
PavloSukhinin: Not sure why package-lock was changed

frontend/src/libs/components/badge/badge.tsx
```diff
@@ -0,0 +1,49 @@
...
+}: Properties): JSX.Element => {
+  const isEvent = Boolean(handleClick);
```
PavloSukhinin:
```suggestion
  const hasEvent = Boolean(handleClick);
```

frontend/src/libs/components/badge/badge.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const isColorInPalette = Boolean(paletteColors[color]);
+  const backgroundColor = isColorInPalette ? color : defaultColor;
```
PavloSukhinin:
```suggestion
  const backgroundColor = paletteColors[color] ?? defaultColor;
```
to simplify it a bit

frontend/src/libs/components/badge/badge.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const isBackgroundDark = Boolean(darkColors[backgroundColor]);
+  const isClassName = Boolean(className);
```
PavloSukhinin:
```suggestion
  const hasClassName = Boolean(className);
```

PavloSukhinin (changes requested):

frontend/src/libs/components/badge/badge.tsx
albsbz: fixed

package-lock.json
albsbz: removed

frontend/src/libs/components/badge/badge.tsx
albsbz: fixed

frontend/src/libs/components/badge/badge.tsx
albsbz: fixed with  `const backgroundColor = paletteColors[color] ? color : defaultColor;`

frontend/src/libs/components/badge/badge.tsx
albsbz: fixed

PavloSukhinin (dismissed):

Druxight (dismissed):

PavloSukhinin (approved):

PR #105, opened 2023-08-27
th-96: Add custom map design and tow-truck icon

h0wter:
Created a custom design for the map and added a tow-truck icon.

map


frontend/src/libs/packages/map/map.package.ts
```diff
@@ -30,6 +32,7 @@ class MapService implements IMapService {
     this.map = new google.maps.Map(mapElement, {
       center,
       zoom,
+      mapId: '2db763871084f8b6',
```
albsbz: If this from google project, maybe better move it to .ENV?

Druxight: use JSON config

Druxight (changes requested):

frontend/src/libs/packages/map/map.package.ts
```diff
@@ -1,6 +1,9 @@
+import truckImg from 'src/assets/img/tow-truck.png';
```
PavloSukhinin:
```suggestion
import truckImg from '~/assets/img/tow-truck.png';
```

frontend/src/libs/packages/map/map.config.json
```diff
@@ -0,0 +1,227 @@
...
+      {
+        "visibility": "off"
+      }
+    ]
+  },
+  {
+    "featureType": "poi.sports_complex",
+    "stylers": [
+      {
+        "visibility": "off"
```
kalinkaaaa14, line 99: quick question - do we need to add such many feature types with visibility off? are they enabled by default?

h0wter: The standard map shows all places, captions. To match the design I had to turn off various points of interest. I turned off only what does not match the design.

kalinkaaaa14: ok, got it, thank you

kalinkaaaa14 (approved):

DmytroVoronkov (approved):

Druxight (approved):

PR #107, opened 2023-08-29
th-97: Implement mail service

mV22110702:


backend/src/libs/packages/mailer/mailer.package.ts
```diff
@@ -0,0 +1,77 @@
+import { type Transporter } from 'nodemailer';
+import { createTransport } from 'nodemailer';
```
kalinkaaaa14: unite imports

backend/src/libs/packages/mailer/mailer.package.ts
```diff
@@ -0,0 +1,77 @@
...
+import { type Options } from 'nodemailer/lib/mailer';
+
+import { MailerConnectionError } from '~/libs/exceptions/mailer/mailer-connection.exception.js';
```
kalinkaaaa14:
```suggestion
import { MailerConnectionError } from '~/libs/exceptions/mailer/mailer.js';
```

shared/src/libs/exceptions/config-validation-error/config-validation-error.exception.ts
```diff
@@ -0,0 +1,14 @@
+type Constructor = {
+  message: string;
+  cause?: unknown;
+};
```
kalinkaaaa14: maybe let's create a separate type ErrorConstructor, and reuse it in all places where such type is now created ( I mean in every separate exception now )

backend/src/packages/mail/mail.service.ts
```diff
@@ -0,0 +1,32 @@
+import { type Mailer } from '~/libs/packages/mailer/mailer.package';
```
kalinkaaaa14:
```suggestion
import { type Mailer } from '~/libs/packages/packages.js';
```

backend/src/packages/mail/libs/views/views-mapper.ts
```diff
@@ -0,0 +1,8 @@
+import { TemplateName } from '../enums/enums.js';
+import { plainView } from './plain/plain.js';
+
+const ViewsMapper = {
```
kalinkaaaa14: as it is a map, you should create libs/maps/maps, libs/maps/templateNameToView for example
and don't forget to add a type for this map

backend/src/packages/mail/libs/types/template-names.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type ValueOf } from '~/libs/types/types.js';
+
+import { type TemplateName } from '../enums/enums.js';
+
+type TemplateNames = ValueOf<typeof TemplateName>;
```
kalinkaaaa14:
```suggestion
type TemplateNameValues = ValueOf<typeof TemplateName>;
```

backend/src/libs/packages/config/config.package.ts
```diff
@@ -28,6 +30,28 @@ class Config implements IConfig {
...
   private get envSchema(): TConfig<EnvironmentSchema> {
+    convict.addFormat({
+      name: 'boolean',
+      validate: (value: string) => {
+        if (typeof value !== 'boolean') {
+          throw new ConfigValidationError({
+            message: 'Invalid SMTP_TLS format',
+          });
+        }
+      },
```
PavloSukhinin: not sure about message, maybe we'll re-use this format for other env variables

backend/src/libs/packages/config/config.package.ts
```diff
@@ -28,6 +30,28 @@ class Config implements IConfig {
...
+    });
+
+    convict.addFormat({
+      name: 'email',
+      validate: (value: string) => {
+        if (!FormatRegex.EMAIL.test(value)) {
+          throw new ConfigValidationError({
+            message: 'Invalid Sendgrid sender email format',
```
PavloSukhinin: same as above

backend/src/packages/mail/libs/maps/template-name-to-view.map.ts
```diff
@@ -0,0 +1,11 @@
+import { type IView } from '~/libs/packages/packages.js';
+
+import { TemplateName } from '../enums/enums.js';
+import { type TemplateNameValues } from '../types/types.js';
+import { plainView } from '../views/views.js';
+
+const TemplateNameToView: Record<TemplateNameValues, IView> = {
```
kalinkaaaa14:
```suggestion
const templateNameToView: Record<TemplateNameValues, IView> = {
```

shared/src/libs/types/require-property.ts
```diff
@@ -0,0 +1,4 @@
+type RequireProperty<T extends object, K extends keyof T = keyof T> = T &
+  Pick<T, K>;
```
PavloSukhinin:
```suggestion
type RequireProperty<T extends object, K extends keyof T = keyof T> = T &
  Required<Pick<T, K>>;
```
otherwise it won't do anything 🤔

backend/src/libs/packages/mailer/mailer.ts
```diff
@@ -0,0 +1,11 @@
+import { config } from '~/libs/packages/config/config.js';
+import { logger } from '~/libs/packages/logger/logger.js';
+
+import { Mailer } from './mailer.package.js';
+
+const mailer = new Mailer({ config, logger });
+
+await mailer.connect();
```
PavloSukhinin: Not sure that we want to drop our server if there is something with email service. It's not the same as a database - an application can work even without it.

PavloSukhinin (changes requested):

kalinkaaaa14 (dismissed):

kalinkaaaa14 (approved):

PavloSukhinin (approved):

PR #111, opened 2023-08-29
th-16: Implement jwt auth

Some14u:
It is a draft pr, what is left to implement is the way to apply auth strategies to controller enpoints. I'm going to slightly modify controller core code.


backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { type JwtPayload } from 'shared/build';
+
+const assertJwtPayload = (payload: unknown): payload is JwtPayload => {
+  return typeof payload === 'object' && !!payload && ('id' in payload) && typeof payload.id === 'number';
```
mV22110702: Can we make use of joi here?

Some14u: Yes, I think so

backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -39,6 +41,17 @@ const groups = pgTable('groups', {
   updatedAt: timestamp('updated_at').notNull().defaultNow(),
 });
 
+const groupsRelations = relations(groups, ({ many }) => ({
```
PavloSukhinin: Not sure that we'll use it

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  fastify.decorate(AuthStrategy.VERIFY_JWT, async (request: FastifyRequest, _: FastifyReply, done: DoneFuncWithErrOrRes): Promise<void> => {
+    try {
+      const token = request.headers.authorization?.replace('Bearer ', '');
+
+      if (!token) {
+        return done(createUnauthorizedError(HttpMessage.UNAUTHORIZED));
+      }
+
+      const user = await userService.findByAccessToken(token);
```
PavloSukhinin: Maybe we should decode token, find user by id and then compare tokens. 
- lookup by PK is faster
- id is unique, we can ensure that result will be exatly 1

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,92 @@
...
+declare module '@fastify/jwt' {
+  interface FastifyJWT {
+    payload: { id: UserEntityT['id'] }
+    user: UserEntityObjectWithGroupT;
+  }
+}
+const authPlugin = fp<AuthPluginOptions>(async (fastify, options) => {
+  const { userService, config, encryptService } = options;
+
+  await fastify.register(jwt, { secret: config.ENV.JWT.SECRET });
```
PavloSukhinin: Not sure that we'll need auth strategies, can we just use 1 plugin? Something like this
```
type UnprotectedRoute {
  routerPath: string;
  methods: HttpMethod[]
}

type AuthorizationPluginParameters = {
  unprotectedRoutesConfig: UnprotectedRoute[];
  userService: UserService;
  token: TypeOfTokenPackage;
};

const authorization = fp(
  (
    fastify: FastifyInstance,
    { unprotectedRoutesConfig, userService, token }: AuthPluginParameters,
  ) => {
    fastify.decorateRequest('user', null);

    fastify.addHook('onRequest', async (request: FastifyRequest) => {
      const { headers, method, routerPath } = request;

      const isUnprotectedRoute = unprotectedRoutesConfig.some((whiteRoute) => {
        const isUnprotectedPath = unprotectedRoutesConfig.routerPath === routerPath;
        const isAllowedMethod = unprotectedRoutesConfig.methods.includes(
          method
        );

        return isUnprotectedRoute && isAllowedMethod;
      });

      if (isUnprotectedRoute) {
        return;
      }

      const requestToken = headers.authorization?.replace('Bearer ', '');

      if (!requestToken) {
        // throw error here
      }

      const { id: userId } = token.decode<UserTokenPayload>(requestToken);

      if (!userId) {
        // throw an error
      }

      const user = await userService.findById(userId);

      if (!user || token !== user.accessToken) {
        // throw an error
      }

      request.user = user;
    });
  }
);
```

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,101 @@
+import { type FastifyAuthFunction } from '@fastify/auth';
```
PavloSukhinin: Let's create separate Token package that will have 2 methods `create` or `sign` and `decode`
`backend/src/libs/packages/token`
We might use it not only for user auth

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
Some14u: However i doubt Joi will cast the value to required type as this method does.

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,101 @@
...
+import { assertJwtPayload } from './libs/helpers/assert-jwt-payload.helper.js';
+import { createUnauthorizedError } from './libs/helpers/create-unauthorized-error.helper.js';
+import { type AuthPluginOptions } from './libs/types/auth-plugin-options.type.js';
+
+declare module '@fastify/jwt' {
+  interface FastifyJWT {
+    payload: { id: UserEntityT['id'] }
+    user: UserEntityObjectWithGroupT;
+  }
+}
```
PavloSukhinin: Not sure why we need this

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      if (!user) {
+        return done(createUnauthorizedError(HttpMessage.WRONG_EMAIL));
+      }
+
+      const passwordsAreEqual = await encryptService.compare(password, user.passwordHash);
+
+      if (!passwordsAreEqual) {
+        return done(createUnauthorizedError(HttpMessage.WRONG_PASSWORD));
+      }
```
PavloSukhinin: I believe all this should be performed in auth.service by signIn method

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,101 @@
+import { type FastifyAuthFunction } from '@fastify/auth';
+import jwt from '@fastify/jwt';
+import { type FastifyReply, type FastifyRequest } from 'fastify';
+import fp from 'fastify-plugin';
+import { type UserEntityT, type UserSignInRequestDto, HttpMessage } from 'shared/build/index.js';
```
PavloSukhinin: Import this from corresponding directories on backend e.g `~/libs/enums/enums.ts`

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
Some14u: @mV22110702 the problem here is that this function narrows down the type to object, having id. Joi can check the value on javascript level, but on the typescript the value would still be unknown.
Although, there is a workaround for this. I can define the type of jwt payload globally making jwt verificator think that the type contains id by default. Then it will be convenient to check it as you suggested using Joi.

backend/src/libs/packages/controller/libs/types/auth-strategy-handler.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type FastifyAuthFunction } from '@fastify/auth';
+import { type FastifyInstance } from 'fastify';
+import { type ValueOf } from 'shared/build';
```
PavloSukhinin:
```suggestion
import { type ValueOf } from '~/libs/types/types.js';
```

backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
```diff
@@ -1,10 +1,8 @@
-import { type UserMocked } from './user-mocked.type.js';
-
 type DefaultApiHandlerOptions = {
   body?: unknown;
   query?: unknown;
   params?: unknown;
-  user?: UserMocked;
+  user?: unknown;
```
PavloSukhinin:
```suggestion
  user?: UserWithGroupT | null;
```

backend/src/packages/auth/auth.controller.ts
```diff
@@ -1,3 +1,6 @@
+import { type UserGroupKey, type UserSignInRequestDto, type ValueOf } from 'shared/build/index.js';
+import { userSignInValidationSchema } from 'shared/build/index.js';
```
PavloSukhinin: Please import from `~/libs/types/types.js`/ `~/libs/enums/enums.js` /etc...

backend/src/packages/auth/auth.service.ts
```diff
@@ -1,43 +1,51 @@
-import { HttpCode, HttpMessage } from '~/libs/enums/enums.js';
+import { type UserGroupKey, type ValueOf } from 'shared/build';
+import { HttpCode, HttpMessage } from 'shared/build';
```
PavloSukhinin: Please, update imports

backend/src/packages/auth/auth.service.ts
```diff
@@ -1,43 +1,51 @@
...
+  public constructor(userService: UserService, groupService: GroupService) {
+    this.userService = userService;
+    this.groupService = groupService;
   }
 
-  public async signUp(
-    userRequestDto: CustomerSignUpRequestDto,
-  ): Promise<CustomerSignUpResponseDto> {
-    const isUserExist = await this.checkExistingUser(userRequestDto);
+  public async signUp(groupName: ValueOf<typeof UserGroupKey>,
```
PavloSukhinin:
```suggestion
  public async signUp(groupKey: ValueOf<typeof UserGroupKey>,
```

backend/src/packages/auth/auth.service.ts
```diff
@@ -1,43 +1,51 @@
...
+    payload: CustomerSignUpRequestDto,
+  ): Promise<UserEntityObjectWithGroupT> {
+    const user = await this.userService.findByEmail(payload.email);
 
-    if (isUserExist) {
+    if (user) {
```
PavloSukhinin, line 62: Not sure why you removed checkIsExistingUser

backend/src/packages/auth/auth.service.ts
```diff
@@ -1,43 +1,51 @@
...
-    return await this.userService.create(userRequestDto);
+    const group = await this.groupService.findByKey(groupName);
+
+    if (!group) {
+      throw new HttpError({
+        message: HttpMessage.INVALID_GROUP,
+        status: HttpCode.BAD_REQUEST
+      });
+    }
+    const result = await this.userService.create({ ...payload, groupId: group.id });
```
PavloSukhinin: Let's generate access token in sign up method

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
PavloSukhinin: Can't see type cast here

backend/src/packages/auth/libs/helpers/create-unauthorized-error.helper.ts
```diff
@@ -0,0 +1,9 @@
+import { HttpCode, HttpError } from 'shared/build';
```
PavloSukhinin: Please use imports from ~/libs/

backend/src/packages/auth/libs/types/auth-plugin-options.type.ts
```diff
@@ -0,0 +1,11 @@
+import { type IConfig } from '~/libs/packages/config/config.js';
+import { type EncryptService } from '~/libs/packages/encrypt/encrypt.package.js';
+import { type UserService } from '~/packages/users/user.service';
```
PavloSukhinin: missing file extension

backend/src/packages/business/business.repository.ts
```diff
@@ -23,14 +23,18 @@ class BusinessRepository implements IRepository {
...
-      .driver()
-      .select()
-      .from(this.businessSchema)
-      .where(eq(this.businessSchema.id, id));
+  public find(partial: Partial<BusinessEntityT>):
+    ReturnType<IRepository<BusinessEntityT>['find']> {
+    const queries = Object.entries(partial).map(([key, value]) => eq(
+      this.businessSchema[key as keyof typeof partial],
+      value as NonNullable<typeof value>
+    ));
```
PavloSukhinin: I thought we agreed that we will implement specific find methods when we'll need them, instead of one generic method that can search by anything. Because right now it's quite ambiguous that it uses the AND clause.

backend/src/packages/business/business.service.ts
```diff
@@ -20,12 +20,12 @@ class BusinessService implements IService {
...
-  public async find(
+  public async findById(
     id: number,
-  ): Promise<OperationResult<BusinessEntityT | null>> {
-    const business = await this.businessRepository.find(id);
+  ): Promise<BusinessEntityT | null> {
+    const result = await this.businessRepository.find({ id });
 
-    return { result: business ? business.toObject() : null };
+    return result.length === 1 ? BusinessEntity.initialize(result[0]).toObject() : null;
```
PavloSukhinin: const [business = null] = await 

return business ? BusinessEntity.initialize(business).toObject() : null;

backend/src/packages/groups/group.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  public constructor(groupRepository: GroupRepository) {
+    this.groupRepository = groupRepository;
+  }
+
+  public async findByName(name: GroupEntityT['name']): Promise<GroupEntityObjectT | null> {
+    const result = await this.groupRepository.find({ name });
+
+    return result.length === 1 ? GroupEntity.initialize(result[0]).toObject() : null;
+  }
```
PavloSukhinin: I believe we shouldn't look up for a group by it's name, only by key.
Also avoid magic numbers, prefer destructuring

```const [group = null] = await ;
return group ? GroupEntity...
```

backend/src/packages/users/libs/types/user-models.type.ts
```diff
@@ -0,0 +1,30 @@
+import { type InferModel } from 'drizzle-orm';
+import { type UserEntityT } from 'shared/build/index.js';
```
PavloSukhinin: from '~/libs/types/'

backend/src/packages/users/libs/types/user-models.type.ts
```diff
@@ -0,0 +1,30 @@
...
+type UserEntityObjectWithGroupT = (UserEntityObjectT & { group: GroupEntityObjectT }) | null;
+
+export {
+  type UserDatabaseModel,
+  type UserDatabaseModelCreateUpdate,
+  type UserDatabaseModelWithGroup,
+  type UserEntityObjectT,
+  type UserEntityObjectWithGroupT
+};
+export { type UserEntityCreateUpdate };
```
PavloSukhinin: merge exports

backend/src/packages/users/user.service.ts
```diff
@@ -1,70 +1,126 @@
...
-    return this.userRepository.findByPhone(value);
+  public async findByPhone(phone: UserEntityT['phone']): Promise<UserEntityObjectWithGroupT> {
+    const result = await this.userRepository.find({ phone });
+
+    if (result.length === 1) {
+      return {
+        ...UserEntity.initialize(result[0]).toObject(),
+        group: GroupEntity.initialize(result[0].group).toObject()
+      };
+    }
```
PavloSukhinin: A bit clearer IMO
```suggestion
    const [user = null] = await this.userRepository.find({ phone });
    
    if (!user) {
       return null
    }
    
    const { group, ...goodName } = user;

    return {
         ...UserEntity.initialize(goodName).toObject(),
        group: GroupEntity.initialize(group).toObject()
    }
}
```

backend/src/packages/users/user.service.ts
```diff
@@ -1,70 +1,126 @@
...
 
-  public findById(id: number): ReturnType<IService['find']> {
-    return this.userRepository.findById(id);
+  public async findByEmail(email: UserEntityT['email']): Promise<UserEntityObjectWithGroupT> {
+    const result = await this.userRepository.find({ email });
+
+    if (result.length === 1) {
+      return {
+        ...UserEntity.initialize(result[0]).toObject(),
+        group: GroupEntity.initialize(result[0].group).toObject()
```
PavloSukhinin: same about magic numbers

backend/src/packages/users/user.service.ts
```diff
@@ -1,70 +1,126 @@
...
+      };
+    }
+
+    return null;
+  }
+
+  public async findByEmailRaw(email: UserEntityT['email']): Promise<UserDatabaseModelWithGroup | null> {
+    const result = await this.userRepository.find({ email });
+
+    return result.length === 1 ? result[0] : null;
```
PavloSukhinin: magic numbers

backend/src/packages/users/user.service.ts
```diff
@@ -1,70 +1,126 @@
...
+  }
+
+  public async findByAccessToken(accessToken: NonNullable<UserEntityT['accessToken']>): Promise<UserEntityObjectWithGroupT> {
+    const result = await this.userRepository.find({ accessToken });
+
+    if (result.length === 1) {
+      return {
+        ...UserEntity.initialize(result[0]).toObject(),
+        group: GroupEntity.initialize(result[0].group).toObject()
+      };
```
PavloSukhinin: please, avoid magic numbers

PavloSukhinin (changes requested):
Let's substitute fastify/jwt with [jose](https://www.npmjs.com/package/jose) since we want to use it as separate token service.

backend/src/packages/business/business.repository.ts
Some14u: @PavloSukhinin it uses, but as you can see, only if there are more than one options being passed. For a singular option it has a special case without and.

Some14u: Although, We agreed about findById for service

backend/src/libs/packages/database/schema/tables-schema.ts
Some14u: To estabilish relation requests. It does not work without this.

backend/src/packages/auth/auth.service.ts
Some14u: Right, i changed the method, but forgot about the prop name.

backend/src/packages/auth/auth.service.ts
Some14u: Basically I did my own userService from scratch, and then failed to join because after merge with the recent update there were too many discrepancies. I decided to replace this module entirely, and most likely lost the checkIsExistingUser

backend/src/packages/auth/auth.app-plugin.ts
Some14u: I've found this layout in a couple lessons. I think, the idea is to keep authentication logic in one place.

It may seem logical when you have more than two or three authentication strategies.

backend/src/packages/auth/auth.service.ts
Some14u: Um... yeah, why not.

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
Some14u: This is not strict casting, this called narrowing, and it is in the return type:
... : payload is JwtPayload

backend/src/packages/business/business.service.ts
Some14u: nice. I wondered, why is it not working. It is because of default null.

backend/src/packages/business/business.repository.ts
Some14u: Currently i'm using it in my user service.
Do you want me to make it user-repository specific (as long as i need it there), and make findById default for all other (by putting it into IRepository)?

PavloSukhinin: Let's leave it like this for now and discuss with the team. We'll have separate task to change it if required

backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
```diff
@@ -1,10 +1,10 @@
-import { type UserMocked } from './user-mocked.type.js';
+import { type UserEntityObjectWithGroupT } from '~/packages/users/libs/types/user-models.type';
```
PavloSukhinin:
```suggestion
import { type UserEntityObjectWithGroupT } from '~/packages/users/users.js';
```

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,67 @@
...
+
+declare module 'fastify' {
+  interface FastifyInstance {
+    [AuthStrategy.VERIFY_JWT]: FastifyAuthFunction;
+  }
+
+  interface FastifyRequest {
+    user: UserEntityObjectWithGroupT;
+  }
+}
```
PavloSukhinin: Create `backend/src/index.d.ts` and move this declaration there. Also update `backend/src/tsconfig.json` - add `"files": ["src/index.d.ts"]`

backend/src/libs/packages/database/schema/tables-schema.ts
PavloSukhinin: I meant only `groupsRelations` - I believe everything will work fine without it

backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -39,6 +41,17 @@ const groups = pgTable('groups', {
...
+  users: many(users),
+}));
+
+const usersRelations = relations(users, ({ one }) => ({
+  groups: one(groups, {
```
PavloSukhinin: It's one group, so let's name it `group`

backend/src/libs/packages/server-application/server-app.ts
```diff
@@ -110,6 +133,8 @@ class ServerApp implements IServerApp {
   private initErrorHandler(): void {
     this.app.setErrorHandler(
       (error: FastifyError | ValidationError, _request, replay) => {
+        // console.error(error);
```
PavloSukhinin: remove it

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const authPlugin = fp<AuthPluginOptions>((fastify, options, done) => {
+  const { userService, jwtService } = options;
+
```
PavloSukhinin, line 13: Let's add `fastify.decorateRequest('user', null)`?
https://fastify.dev/docs/v2.15.x/Documentation/Decorators/

backend/src/packages/auth/auth.controller.ts
```diff
@@ -6,13 +6,20 @@ import {
...
   customerSignUpValidationSchema,
 } from '~/packages/users/users.js';
 
 import { type AuthService } from './auth.service.js';
 import { AuthApiPath } from './libs/enums/enums.js';
+import {
+  type UserGroupKey,
+  type UserSignInRequestDto,
+} from './libs/types/types.js';
+import { userSignInValidationSchema } from './libs/validation-schemas/user-sign-in.validation-schema.js';
```
PavloSukhinin:
```suggestion
import { userSignInValidationSchema } from './libs/validation-schemas/validation-schemas.js';
```

backend/src/packages/auth/auth.service.ts
```diff
@@ -1,42 +1,129 @@
 import { HttpCode, HttpMessage } from '~/libs/enums/enums.js';
 import { HttpError } from '~/libs/exceptions/exceptions.js';
+import { type IEncryptService } from '~/libs/interfaces/encrypt.interface.js';
+import { type IJwtService } from '~/libs/interfaces/jwt.inteface.js';
```
PavloSukhinin: from '~/libs/interfaces/interfaces.js'

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
```diff
@@ -0,0 +1,12 @@
+import { type JwtPayload } from 'shared/build/index.js';
+
+const assertJwtPayload = (payload: unknown): payload is JwtPayload => {
+  return (
+    typeof payload === 'object' &&
+    !!payload &&
+    'id' in payload &&
+    typeof payload.id === 'number'
```
PavloSukhinin: So why not joi?

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
```diff
@@ -0,0 +1,12 @@
+import { type JwtPayload } from 'shared/build/index.js';
```
PavloSukhinin: `~libs/packages/jwt/jwt.js`

backend/src/packages/groups/group.repository.ts
```diff
@@ -0,0 +1,84 @@
+import { and, eq } from 'drizzle-orm';
+
+import { type IRepository } from '~/libs/interfaces/interfaces.js';
+import { type IDatabase } from '~/libs/packages/database/database.js';
+import { type DatabaseSchema } from '~/libs/packages/database/schema/schema.js';
```
PavloSukhinin: let's export `type DatabaseSchema` from `/src/libs/packages/database/database.ts` and import it as 
`import { type DatabaseSchema, type IDatabase } from '~/libs/packages/database/database.js';`

backend/src/packages/groups/libs/types/enums/user-group.key.enum.ts
```diff
@@ -0,0 +1 @@
+export { UserGroupKey } from 'shared/build/index.js';
```
PavloSukhinin: `backend/src/packages/groups/libs/enums/user-group.key.enum.ts` looks like redundant `types` dir

backend/src/packages/users/user.service.ts
```diff
@@ -1,70 +1,136 @@
...
 
-  public findByPhone(value: string): Promise<UserEntity | null> {
-    return this.userRepository.findByPhone(value);
+    return result.map((item) => UserEntity.initialize(item).toObject());
   }
 
-  public findById(id: number): ReturnType<IService['find']> {
-    return this.userRepository.findById(id);
+  // TODO: Do we need this?
+  public async findByPhone(
```
PavloSukhinin: Let's keep it and remove TODO comment

PavloSukhinin (changes requested):

shared/src/libs/packages/http/libs/enums/http-error-message.enum.ts
```diff
@@ -1,9 +1,15 @@
+/* eslint-disable @typescript-eslint/quotes */
 const HttpMessage = {
   USER_EXISTS: 'User already exists',
   BUSINESS_ALREADY_EXISTS: 'Owner already has business!',
   NAME_ALREADY_REGISTERED: 'Business with such name already exists!',
   INVALID_USER_GROUP: 'User of the group cannot create business!',
   BUSINESS_DOES_NOT_EXIST: 'Business does not exist!',
+  UNAUTHORIZED: "You're not authorized",
```
PavloSukhinin: Let's use `'You are....'` and enable eslint

PavloSukhinin (changes requested):

backend/src/packages/auth/auth.app-plugin.ts
Some14u: Ah, yes, it was impossible before because of @fastify/jwt, which has been doing this automatically.

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
Some14u: because after using joi I would have to use "as" to forcefully cast the type. But if you have nothing against it, I will redo.

shared/src/libs/packages/http/libs/enums/http-error-message.enum.ts
Some14u: :D

backend/src/packages/auth/libs/helpers/assert-jwt-payload.helper.ts
Some14u: Forget it, I've added Joi

backend/src/libs/packages/database/schema/tables-schema.ts
Some14u: I didn't know it is possible, thought "groups" here stands for database name.

backend/src/packages/auth/libs/enums/enums.ts
```diff
@@ -1 +1,3 @@
+export { AuthStrategy } from './auth-strategy.enum.js';
 export { AuthApiPath } from 'shared/build/index.js';
+export { UserGroupKey } from 'shared/build/index.js';
```
PavloSukhinin:
```suggestion
export { AuthApiPath, UserGroupKey } from 'shared/build/index.js';
```

backend/src/packages/auth/auth.app-plugin.ts
```diff
@@ -0,0 +1,57 @@
+import { type FastifyReply, type FastifyRequest } from 'fastify';
+import fp from 'fastify-plugin';
+import { jwtPayloadSchema } from 'shared/build/index.js';
```
PavloSukhinin:
```suggestion
import fp from 'fastify-plugin';
import { jwtPayloadSchema } from './libs/validation-schemas/validation-schemas.js';
```

PavloSukhinin (changes requested):

backend/src/libs/packages/database/schema/tables-schema.ts
Some14u: I see now, okay, i'll remove it

PavloSukhinin (approved):

PR #114, opened 2023-08-29
th-112: Add query parameters service

h0wter:
All functions work with either single or multiple arguments.
`getQueryParameters('param')` or `getQueryParameters('param', 'param1')`
`removeQueryParameters('param')` or `removeQueryParameters('param', 'param1')`
`setQueryParameters({param: value})` or `setQueryParameters({param: value, param1: value1})`
`setQueryParameters` can take both a string and a number as a property value.


frontend/src/libs/hooks/use-query-parameters/use-query-parameters.hook.ts
```diff
@@ -0,0 +1,63 @@
+import { useCallback, useSearchParameters } from '../hooks.js';
```
PavloSukhinin: Circular dependency, be careful with that

Druxight (dismissed):

frontend/src/libs/hooks/use-query-parameters/use-query-parameters.hook.ts
```diff
@@ -0,0 +1,63 @@
...
+
+type Queries = Record<string, string | null>;
+
+type Parameters = Record<string, string | number>;
```
PavloSukhinin: rename this, Parameters is build-in TS type

PavloSukhinin (changes requested):

DmytroVoronkov (approved):

frontend/src/libs/hooks/use-query-parameters/use-query-parameters.hook.ts
```diff
@@ -0,0 +1,64 @@
...
+      setSearchParameters(newSearchParameters);
+    },
+    [searchParameters, setSearchParameters],
+  );
+
+  const removeQueryParameters = useCallback(
+    (...keys: string[]): void => {
+      const newSearchParameters = new URLSearchParams(searchParameters);
+
+      if (keys.length > 1) {
```
Some14u, line 48: consider when there is no parameters being passed. I think you can add "remove everything" for that particular case.

Some14u (approved):
Simple and clean. I like it, however in my opinion it lacks possibility to pass and handle default values.
For example, if i need some key value, which has it's own default, i'd expect this to return the default value when there is no corresponding key in the query.

PavloSukhinin (approved):

kalinkaaaa14 (approved):

PR #124, opened 2023-08-31
th-85: Business sign up flow

TetianaDoroshko:
Business & Customer sign up
Notes:
1. It contain a migration from the previous PR, which is deprecated.

2. In BusinessSignUpResponceDto I've created temporary type for GroupEntity because the types used in groupService are not  available it Shared scope.
```typescript
type GroupEntityT = { id: number; name: string; key: string };

type BusinessSignUpResponseDto = Omit<
  UserEntityT,
  'passwordHash' | 'passwordSalt' | 'groupId'
> & { group: GroupEntityT } & { business: BusinessEntityT[] };
``` 
If everythnig else is well, I'll ask Max to do with it something.


backend/src/packages/auth/auth.service.ts
```diff
@@ -37,36 +40,52 @@ class AuthService {
...
+    return Boolean(existingUser);
+  }
+
+  public async signUpCustomer(
     payload: CustomerSignUpRequestDto,
   ): Promise<UserEntityObjectWithGroupT> {
-    const user = await this.userService.findByEmail(payload.email);
+    const isExistingUser = await this.checkIsExistingUser(payload);
+
+    // const { phone, email, firstName, lastName, password } = payload;
```
Druxight: remove it

backend/src/packages/auth/auth.service.ts
```diff
@@ -87,6 +106,61 @@ class AuthService {
...
+    const userWithToken = await this.generateAccessTokenAndUpdateUser(
+      newUser.id,
+    );
+
+    const business = await this.businessService.create({
+      payload: {
+        companyName,
+        taxNumber,
+      },
+      owner: { ...newUser, group },
```
Druxight, line 156: do we need to pass group here ?

frontend/src/pages/auth/auth.tsx
```diff
@@ -18,10 +21,22 @@ import styles from './styles.module.css';
...
+  const navigate = useNavigate();
 
   const location: Location = useLocation();
   const mode = location.state as ValueOf<typeof AuthMode>;
 
+  const { dataStatus } = useAppSelector(({ auth }) => ({
+    dataStatus: auth.dataStatus,
+  }));
+
+  // TEMPORARY!
```
Druxight: In such cases its better to leave //TODO comments: 

```
//TODO refactor after {something} is ready
```

shared/src/packages/users/libs/types/business-sign-up-response-dto.type.ts
```diff
@@ -1,5 +1,15 @@
+import { type BusinessEntityT } from '~/packages/business/business.js';
+
 import { type UserEntityT } from './user-entity.type.js';
 
-type BusinessSignUpResponseDto = Pick<UserEntityT, 'id' | 'phone'>;
+// import { type UserGroupEntityT } from './user-group-entity.type.js';
+
+//temporary type!!!
```
Druxight: why ?

Druxight (changes requested):

shared/src/packages/users/libs/types/business-sign-up-response-dto.type.ts
TetianaDoroshko: in Shared scope we have type UserGroupEntityT = { id: number; name: UserGroupNameT; key: UserGroupKeyT; }
but it is not suitable, because groupService returns in type GroupEntityObjectT =  { id: number; name: string; key: string; }. But I can't use it there because it exist in backend only and it is impossible to import it in shared.
I thought that use UserGroupEntityT is better to use and even started to correct, but there are a lot of changes to do in someone's code.
So I desided to pull this for now, and then deal with this type.

backend/src/packages/auth/auth.service.ts
TetianaDoroshko: the businessService.create method requires so...

backend/src/packages/auth/auth.controller.ts
```diff
@@ -33,10 +36,23 @@ class AuthController extends Controller {
...
-        this.signUp(
+        this.signUpCustomer(
           options as ApiHandlerOptions<{
             body: CustomerSignUpRequestDto;
-            params: { groupName: ValueOf<typeof UserGroupKey> };
+          }>,
+        ),
+    });
+
+    this.addRoute({
```
PavloSukhinin, line 46: Please add swagger docs for new route

backend/src/libs/packages/database/generated-schema/0004_email_unique_index.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX IF NOT EXISTS "users_email_unique_idx" ON "users" ("email");
```
PavloSukhinin, line 1: don't forget to update schema in readme.md

backend/src/packages/auth/auth.service.ts
```diff
@@ -87,6 +106,61 @@ class AuthService {
...
+      email,
+      firstName,
+      lastName,
+      password,
+      groupId: group.id,
+    });
+
+    const userWithToken = await this.generateAccessTokenAndUpdateUser(
+      newUser.id,
+    );
```
PavloSukhinin, line 149: can we create access token and then create user to reduce db calls amount?

backend/src/packages/users/libs/types/user-models.type.ts
```diff
@@ -33,11 +34,16 @@ type UserEntityObjectWithGroupT = UserEntityObjectT & {
   group: GroupEntityObjectT;
 };
 
+type UserEntityObjectWithGroupAndBusinessT = UserEntityObjectWithGroupT & {
+  business: BusinessEntityT;
+};
```
PavloSukhinin: I believe this type should be declared in shared module

frontend/src/libs/helpers/get-error-message.ts
```diff
@@ -0,0 +1,9 @@
+const getErrorMessage = (error: unknown): string => {
```
PavloSukhinin: let's create a dir for it and add .helper suffix to file name.
`frontend/src/libs/helpers/get-error-message/get-error-message.helper.ts`

frontend/src/slices/auth/actions.ts
```diff
@@ -1,6 +1,7 @@
 import { createAsyncThunk } from '@reduxjs/toolkit';
 
 import { type AuthMode } from '~/libs/enums/enums.js';
+import { getErrorMessage } from '~/libs/helpers/get-error-message.js';
```
PavloSukhinin:
```suggestion
import { getErrorMessage } from '~/libs/helpers/helpers.js';
```

frontend/src/slices/auth/actions.ts
```diff
@@ -14,16 +15,20 @@ import {
 import { name as sliceName } from './auth.slice.js';
 
 const signUp = createAsyncThunk<
-  CustomerSignUpResponseDto | BusinessSignUpResponseDto,
+  CustomerSignUpResponseDto | BusinessSignUpResponseDto | undefined,
```
PavloSukhinin:
```suggestion
  CustomerSignUpResponseDto | BusinessSignUpResponseDto
 ,
```

frontend/src/slices/auth/actions.ts
```diff
@@ -14,16 +15,20 @@ import {
...
-  const { authApi } = extra;
+>(`${sliceName}/sign-up`, async ({ payload, mode }, { extra }) => {
+  const { authApi, notification } = extra;
 
-  return authApi.signUp(payload, mode);
+  try {
+    return await authApi.signUp(payload, mode);
+  } catch (error) {
+    notification.warning(getErrorMessage(error));
+  }
```
PavloSukhinin:
```suggestion
  } catch (error: unknown) {
    notification.warning(getErrorMessage(error));
    throw error;
  }
```
Action should be rejected if something went wrong. If we won't throw an error - it will be fulfilled

PavloSukhinin (changes requested):

backend/src/packages/auth/auth.service.ts
TetianaDoroshko: We need to know a userId for token. So we need do it in two steps: create a user and receive an id then create a token and update the user.

PavloSukhinin: Right, my bad, thanks!

shared/src/packages/users/libs/types/business-sign-up-response-dto.type.ts
TetianaDoroshko: upd
I've create type UserGroupEntityObject = { id: number; name: string; key: string; }; in Shared and used it.
But I still think using type  { id: number; name: UserGroupNameT; key: UserGroupKeyT; } would be better. But it needs to make a lot of changes in groupService and businessService.

shared/src/packages/users/libs/types/business-sign-up-response-dto.type.ts
```diff
@@ -1,5 +1,11 @@
+import { type BusinessEntityT } from '~/packages/business/business.js';
+
 import { type UserEntityT } from './user-entity.type.js';
+import { type UserGroupEntityObjectT } from './user-group-entity.type.js';
 
-type BusinessSignUpResponseDto = Pick<UserEntityT, 'id' | 'phone'>;
+type BusinessSignUpResponseDto = Omit<
+  UserEntityT,
+  'passwordHash' | 'passwordSalt' | 'groupId'
+> & { group: UserGroupEntityObjectT } & { business: BusinessEntityT };
```
PavloSukhinin: same as below, can we declare like this `type BusinessSignUpResponseDto = UserEntityObjectWithGroupAndBusiness`?

shared/src/packages/users/libs/types/customer-sign-up-response-dto.type.ts
```diff
@@ -1,8 +1,9 @@
 import { type UserEntityT } from './user-entity.type.js';
+import { type UserGroupEntityObjectT } from './user-group-entity.type.js';
 
-type CustomerSignUpResponseDto = Pick<
+type CustomerSignUpResponseDto = Omit<
   UserEntityT,
-  'id' | 'phone' | 'accessToken'
->;
+  'passwordHash' | 'passwordSalt' | 'groupId'
+> & { group: UserGroupEntityObjectT };
```
PavloSukhinin: Aren't this duplicate of UserEntityObjectWithGroup? can we declare it like `type CustomerSignUpResponseDto = UserEntityObjectWithGroup`

PavloSukhinin (changes requested):

PavloSukhinin (approved):

Druxight (approved):

PR #130, opened 2023-09-01
th-127: The company name field on the sign up business page does not accept digits as valid value

albsbz:
Added digits as posible symbols to company name validation scheme
compname


Druxight:
32 commits ?

PR #134, opened 2023-09-01
th-88: Add driver flow

MaxHopaytsa:
- Create schemas for driver_details
- Add entity and repository for drivers
- Add service for driver
- Add in business controller routes for interaction with drivers
- Add validations 
- Create form for addштп drivers by business
- Add mermaid diagram for driver_details
add-driver
mermaid_driver
**Note:** The form is not yet in the modal window, when it is known where exactly and on which page this form will be placed, immediately on that page and wrap it in a modal window


shared/src/packages/drivers/libs/validation-schemas/driver-update-request-body.validation-schema.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const driverUpdateRequestBody = joi.object<
+  Omit<DriverUpdateRequestDto, 'businessId'>,
+  true
+>({
+  ...commonSignUpRules,
+  driverLicenseNumber: joi
+    .string()
+    .trim()
+    .regex(/^[A-Z]{3} \d{6}$/)
```
albsbz: Noticed that it's already defined in shared/src/packages/users/libs/validation-schemas/enums/user-validation-rule.enum.ts

MaxHopaytsa: Yes, I added it there myself, but it completely slipped my mind. Thank you for noticing.

backend/src/packages/drivers/driver.entity.ts
```diff
@@ -0,0 +1,72 @@
+import { type IEntity } from '~/libs/interfaces/entity.interface.js';
+import { type NullableProperties } from '~/libs/types/types.js';
+
+import { type DriverEntityT } from './libs/types/types.js';
```
Druxight: Please rename the DriverEntityT to the DriverEntity 
If you fit the collision you can always use 'as'
import { type DriverEntity as DriverEntityT } from './libs/types/types.js';

backend/src/packages/drivers/driver.repository.ts
```diff
@@ -0,0 +1,133 @@
...
+      filterClause.push(
+        eq(this.driverSchema.driverLicenseNumber, driverLicenseNumber),
+      );
+    }
+
+    if (userId) {
+      filterClause.push(eq(this.driverSchema.userId, userId));
+    }
+
+    if (filterClause.length === 0) {
```
Druxight, line 73:
```suggestion
    if (!filterClause.length) {
```

backend/src/packages/drivers/driver.service.ts
```diff
@@ -0,0 +1,157 @@
...
+    const driverObject = driver.toObject();
+
+    return { ...user, group, driver: driverObject };
+  }
+
+  public async update({
+    driverId,
+    payload,
+  }: DriverUpdatePayload): Promise<DriverUpdateResponseDto> {
+    const foundDriverById = await this.findById(driverId);
```
Druxight:
```suggestion
    const foundDriver = await this.findById(driverId);
```

backend/src/packages/drivers/driver.repository.ts
MaxHopaytsa: However, linter generates the following error: **"Use `.length === 0` when checking length is zero."**. Maybe I should put this check in a separate variable?

backend/src/libs/packages/database/generated-schema/0004_lame_miracleman.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE "driver_details"
```
PavloSukhinin: Can we merge those 2 files (re-generate migrations)?

backend/src/libs/packages/database/generated-schema/meta/_journal.json
```diff
@@ -26,6 +26,13 @@
     {
       "idx": 3,
       "version": "5",
+      "when": 1693217534139,
+      "tag": "0003_illegal_wither",
```
PavloSukhinin: Let's name it properly, it was decided by team.

shared/src/packages/drivers/libs/types/service/driver-create-payload.type.ts
```diff
@@ -0,0 +1,19 @@
...
+
+type DriverCreatePayload = {
+  payload: Omit<
+    UserEntityT,
+    'id' | 'passwordHash' | 'passwordSalt' | 'groupId' | 'accessToken'
+  > &
+    Pick<DriverEntity, 'driverLicenseNumber'> & {
+      password: string;
+    };
+  groupKey: ValueOf<typeof UserGroupKey>;
```
PavloSukhinin: do we need groupKey here? We know it's driver

frontend/src/pages/business/components/add-driver-form/add-driver-form.tsx
```diff
@@ -0,0 +1,30 @@
...
+import styles from './styles.module.scss';
+
+type Properties = {
+  onSubmit: (payload: DriverAddRequestDto) => void;
+};
+
+const AddDriverForm: React.FC<Properties> = ({ onSubmit }: Properties) => {
+  return (
+    <div className={styles.wrapper}>
+      <h3 className={styles.title}>Add driver</h3>
```
PavloSukhinin: can we re-use some of the styles for typography.scss?

PavloSukhinin (approved):

PavloSukhinin (changes requested):
Accidentally approved it 🥲

backend/src/packages/drivers/driver.service.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    if (!foundDriver) {
+      throw new NotFoundError({});
+    }
+
+    const isDriverDeleted = await this.driverRepository.delete(id);
+
+    if (isDriverDeleted) {
+      return await this.userService.delete(foundDriver.userId);
+    } else {
```
albsbz, line 149: Maybe we should use soft delete for this, because if we delete driver,  we could not see his statistics

backend/src/packages/drivers/driver.service.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    return { ...user, group, driver: driverObject };
+  }
+
+  public async update({
+    driverId,
+    payload,
+  }: DriverUpdatePayload): Promise<DriverUpdateResponseDto> {
+    const foundDriver = await this.findById(driverId);
+
```
albsbz, line 109: Why we make this extra request to DB. If update is not possible db wiil throw error that we can catch

backend/src/packages/drivers/driver.service.ts
```diff
@@ -0,0 +1,157 @@
...
+    const driverObject = driver.toObject();
+
+    return { ...user, driver: driverObject };
+  }
+
+  public async delete(id: number): Promise<boolean> {
+    const foundDriver = await this.findById(id);
+
+    if (!foundDriver) {
+      throw new NotFoundError({});
```
albsbz: As I understand it will throw 500 error, may be should use HttpError

MaxHopaytsa: This is our custom error for Not Found, which has the path ~libs/exceptions/not-found-error. So no, it doesn't throw a 500 error. The only thing that might be worth adding is a different message.

denred: Alex is right, maybe you should use it like this: 
 `throw new HttpError({
        status
        message
      });` In this case, you can't see 500.

backend/src/packages/drivers/driver.service.ts
MaxHopaytsa: I agree, soft deletion is more appropriate here. But for now, I suggest leaving it as it is. Because in addition to changes to the driver repository, we need to do some work in the user repository and the user service. And this task already has too much code. I think it's better to put the implementation of soft delete in a separate task.After all, I don't think that only the driver will need soft deleted.

shared/src/packages/drivers/libs/validation-schemas/driver-update-parameters.validation-schema.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { BusinessValidationMessage } from '~/packages/business/business.js';
+
+import { DriverValidationMessage } from '../enums/enums.js';
+import { type DriverUpdateRequestParameters } from '../types/types.js';
+
+const driverUpdateParameters = joi.object<DriverUpdateRequestParameters, true>({
+  id: joi.number().integer().positive().required().messages({
+    'number': BusinessValidationMessage.ID_MUST_BE_NUMBER,
+  }),
```
PavloSukhinin: can we declare positiveRequiredIntegerSchema and re-use it everywhere?

shared/src/packages/drivers/libs/types/service/driver-update-payload.type.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { type DriverEntity } from '../driver-entity.type.js';
+
+type DriverUpdatePayload = {
+  payload: Omit<
+    UserEntityT,
+    'id' | 'passwordHash' | 'passwordSalt' | 'groupId' | 'accessToken'
+  > & {
+    password?: string;
+  } & Pick<DriverEntity, 'driverLicenseNumber'>;
```
PavloSukhinin: Can we extract this to separate type? Almost duplicates DriverCreatePayload

shared/src/packages/drivers/libs/types/request/driver-update-request-dto.type.ts
```diff
@@ -0,0 +1,12 @@
+import { type UserEntityT } from '~/packages/users/users.js';
+
+import { type DriverEntity } from '../driver-entity.type.js';
+
+type DriverUpdateRequestDto = Omit<
+  UserEntityT,
+  'id' | 'passwordHash' | 'passwordSalt' | 'groupId' | 'accessToken'
+> & {
+  password?: string;
+} & Pick<DriverEntity, 'driverLicenseNumber'>;
```
PavloSukhinin: Can we derive this type from DriverUpdatePayload?

PavloSukhinin (changes requested):

backend/src/packages/business/business.controller.ts
```diff
@@ -200,6 +303,69 @@ class BusinessController extends Controller {
           }>,
         ),
     });
+
+    this.addRoute({
+      path: BusinessApiPath.DRIVERS,
+      method: 'POST',
+      // authStrategy: 'verifyJWT',
```
kalinkaaaa14: why is it commented?

backend/src/packages/business/business.controller.ts
```diff
@@ -396,6 +562,233 @@ class BusinessController extends Controller {
...
+   *                items:
+   *                  $ref: '#/components/schemas/Driver'
+   */
+
+  private async findAllDrivers(
+    options: ApiHandlerOptions<{
+      params: DriverAllByBusinessRequestParameters;
+    }>,
+  ): Promise<ApiHandlerResponse> {
+    const findDrivers = await this.businessService.findAllDriversById(
```
kalinkaaaa14:
```suggestion
    const drivers = await this.businessService.findAllDriversById(
```

shared/src/packages/users/libs/validation-schemas/enums/user-validation-message.enum.ts
```diff
@@ -6,6 +6,8 @@ const UserValidationMessage = {
   PHONE_NOT_VALID: 'Must be 8 - 17 digits, start with +',
   TAX_NUMBER_NOT_VALID: 'Must be 8 or more characters',
   COMPANY_NAME_NOT_VALID: 'Must be 8 or more characters',
+  DRIVER_LICENSE_NUMBER:
```
kalinkaaaa14:
```suggestion
  DRIVER_LICENSE_NUMBER_NOT_VALID:
```

shared/src/packages/drivers/libs/validation-schemas/driver-update-parameters.validation-schema.ts
kalinkaaaa14: agree

shared/src/packages/drivers/libs/validation-schemas/driver-add-request-body.validation-schema.ts
```diff
@@ -0,0 +1,24 @@
...
+  ...commonSignUpRules,
+  driverLicenseNumber: joi
+    .string()
+    .trim()
+    .regex(UserValidationRule.DRIVER_LICENSE_NUMBER)
+    .required()
+    .messages({
+      'string.empty': DriverValidationMessage.DRIVER_LICENSE_NUMBER_REQUIRED,
+    }),
+});
```
kalinkaaaa14: this validation is also duplicated in several places, let's create it once and reuse it ?

shared/src/packages/drivers/libs/types/service/driver-create-payload.type.ts
```diff
@@ -0,0 +1,16 @@
+import { type UserEntityT } from '~/packages/users/libs/types/types.js';
+
+import { type DriverEntity } from '../driver-entity.type.js';
+
+type DriverCreatePayload = {
+  payload: Omit<
+    UserEntityT,
+    'id' | 'passwordHash' | 'passwordSalt' | 'groupId' | 'accessToken'
+  > &
```
kalinkaaaa14: it will be shorter to use Pick instead of Omit

shared/src/packages/drivers/libs/types/request/driver-add-request-dto.type.ts
```diff
@@ -0,0 +1,13 @@
+import { type UserEntityT } from '~/packages/users/users.js';
+
+import { type DriverEntity } from '../driver-entity.type.js';
+
+type DriverAddRequestDto = Omit<
+  UserEntityT,
+  'id' | 'passwordHash' | 'passwordSalt' | 'groupId' | 'accessToken'
```
kalinkaaaa14: same about using Pick

shared/src/packages/drivers/libs/enums/driver-validation-message.enum.ts
```diff
@@ -0,0 +1,9 @@
+const DriverValidationMessage = {
+  DRIVER_LICENSE_NUMBER_REQUIRED: 'Driver license number is required',
+  ID_MUST_BE_NUMBER: 'Driver id must be number',
+  USER_ID_MUST_BE_NUMBER: 'Driver user id must be number',
+  BUSINESS_ID_MUST_BE_NUMBER: 'Driver user id must be number',
+  UPDATE_AT_LEAST_ONE_FIELD: 'Update DTO is empty, nothing to update',
```
kalinkaaaa14: this message will see nontech person, change it, now it's not appropriate

shared/src/packages/business/libs/enums/business-api-path.enum.ts
```diff
@@ -1,6 +1,8 @@
 const BusinessApiPath = {
   ROOT: '/',
   $ID: '/:id',
+  DRIVERS: '/:id/driver',
```
kalinkaaaa14: if this route is for finding all drivers of the business, then /:id/drivers

frontend/.eslintrc.yml
```diff
@@ -23,12 +23,12 @@ rules:
...
 overrides:
   - files:
-    - vite.config.ts
+      - vite.config.ts
     rules:
       import/no-default-export:
         - off
   - files:
-    - vite-env.d.ts
+      - vite-env.d.ts
```
kalinkaaaa14: do you need these spaces ?

backend/src/packages/business/business.service.ts
```diff
@@ -105,6 +119,42 @@ class BusinessService implements IService {
...
+      driverId,
+      payload,
+    });
+  }
+
+  public async findAllDriversById(
+    id: number,
+  ): Promise<DriverGetAllResponseDto> {
+    return await this.driverService.findAllByBusinessId(id);
+  }
```
kalinkaaaa14:
```suggestion
  public findAllDriversByBusinessId(
    id: number,
  ): Promise<DriverGetAllResponseDto> {
    return this.driverService.findAllByBusinessId(id);
  }
```

backend/src/packages/business/business.service.ts
```diff
@@ -105,6 +119,42 @@ class BusinessService implements IService {
...
+
+  public async deleteDriver(driverId: number): Promise<boolean> {
+    return await this.driverService.delete(driverId);
+  }
```
kalinkaaaa14:
```suggestion
  public deleteDriver(driverId: number): Promise<boolean> {
    return this.driverService.delete(driverId);
  }
```

backend/src/packages/business/business.service.ts
```diff
@@ -105,6 +119,42 @@ class BusinessService implements IService {
...
+
+  public async updateDriver({
+    driverId,
+    payload,
+  }: DriverUpdatePayload): Promise<DriverUpdateResponseDto> {
+    return await this.driverService.update({
+      driverId,
+      payload,
+    });
+  }
```
kalinkaaaa14:
```suggestion
  public updateDriver({
    driverId,
    payload,
  }: DriverUpdatePayload): Promise<DriverUpdateResponseDto> {
    return this.driverService.update({
      driverId,
      payload,
    });
  }
```

backend/src/packages/business/business.service.ts
```diff
@@ -105,6 +119,42 @@ class BusinessService implements IService {
...
+    const doesBusinessExist = await this.findById(id);
+
+    if (!doesBusinessExist) {
+      throw new HttpError({
+        status: HttpCode.BAD_REQUEST,
+        message: HttpMessage.BUSINESS_DOES_NOT_EXIST,
+      });
+    }
+
+    return await this.driverService.create({ payload, id });
```
kalinkaaaa14:
```suggestion
    return this.driverService.create({ payload, id });
```

backend/src/packages/drivers/driver.service.ts
```diff
@@ -0,0 +1,158 @@
...
+        message: HttpMessage.DRIVER_DOES_NOT_EXIST,
+        status: HttpCode.NOT_FOUND,
+      });
+    }
+
+    const isDriverDeleted = await this.driverRepository.delete(id);
+
+    if (isDriverDeleted) {
+      return await this.userService.delete(foundDriver.userId);
+    } else {
```
kalinkaaaa14, line 149: actually there can be also problems, for example
you deleted driver, and then try to delete user, but this driver has connections with trucks, so you will need to delete firstly all these connections and the delete user. 
so, I suppose soft delete will be better for these decision, but not in this pr, because it will be too large
leave a FIXME note and then we can create a separate ticket for this purpose ( actually when we will need a driver deletion flow)

kalinkaaaa14 (changes requested):

kalinkaaaa14:
also there are some troubles with er schema I think
1 user can have 0 or 1 driver details ( on your picture it's shown as many)
1 user can have 0 or 1 business ( same, now shown as many )

backend/src/packages/drivers/driver.service.ts
MaxHopaytsa: Yes, we have already discussed this, and we decided to make it a separate task

backend/src/packages/business/business.service.ts
MaxHopaytsa: I agree with all the async/await deletions except this one. Because here we don't just pass the data on to driverService.create, but before that we make a request for the existence of such a business to the business service method findById, which already forces us to use an asynchronous method.

shared/src/packages/drivers/libs/types/service/driver-create-payload.type.ts
MaxHopaytsa: Actually, after checking, Pick in this case gives a bit larger code

shared/src/packages/drivers/libs/validation-schemas/driver-request-body.validation-schema.ts
```diff
@@ -0,0 +1,24 @@
+import joi from 'joi';
+
+import { commonSignUpRules } from '~/packages/users/libs/validation-schemas/common-rules/common-rules.js';
+import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enums/enums.js';
+
+import { DriverValidationMessage } from '../enums/enums.js';
+import { type DriverRequestDto } from '../types/types.js';
+
+const driverRequestBody = joi.object<
```
kalinkaaaa14: let's give it more specific naming like driverAddRequestBody

shared/src/packages/drivers/libs/types/request/driver-request-parameters.type.ts
```diff
@@ -0,0 +1,3 @@
+type DriverRequestParameters = { id: number; driverId: number };
```
kalinkaaaa14: id - it's businessId or userId? maybe let's add more specific naming to understand better

shared/src/libs/validation-schemas/helper-schemas/positive-required-integer-schema.validation-schema.ts
kalinkaaaa14, line 1: validation-schemas/common-schemas

backend/src/packages/business/business.controller.ts
```diff
@@ -206,6 +305,69 @@ class BusinessController extends Controller {
           }>,
         ),
     });
+
+    this.addRoute({
+      path: BusinessApiPath.DRIVERS,
+      method: 'POST',
+      authStrategy: 'verifyJWT',
```
kalinkaaaa14: is it enough to check that this action can be done only by authenticated user with group business?

backend/src/packages/business/business.controller.ts
```diff
@@ -206,6 +305,69 @@ class BusinessController extends Controller {
...
+            body: DriverRequestDto;
+            params: { id: number };
+          }>,
+        ),
+    });
+
+    this.addRoute({
+      path: BusinessApiPath.DRIVER_$ID,
+      method: 'PUT',
+      authStrategy: 'verifyJWT',
```
kalinkaaaa14: same question about updating

shared/src/packages/drivers/libs/types/service/driver-payload.type.ts
```diff
@@ -0,0 +1,16 @@
...
+
+type DriverPayload = {
+  payload: Omit<
+    UserEntityT,
+    'id' | 'passwordHash' | 'passwordSalt' | 'groupId' | 'accessToken'
+  > &
+    Pick<DriverEntity, 'driverLicenseNumber'> & {
+      password: string;
+    };
+  id: number;
```
kalinkaaaa14: to make it clear
```suggestion
  businessId: number;
```

shared/src/packages/drivers/libs/types/service/driver-payload.type.ts
```diff
@@ -0,0 +1,16 @@
+import { type UserEntityT } from '~/packages/users/libs/types/types.js';
+
+import { type DriverEntity } from '../driver-entity.type.js';
+
+type DriverPayload = {
```
kalinkaaaa14: also let's rewrite it, it's quite difficult to read now this type
I propose something like
type UserCommonDetails - pick necessary fields from User
type DriverDetails - pick driver license number
type DriverDto = UserCommonDetails & DriverDetails
type DriverCreateUpdateRequestBody - DriverDto & { password: string }

kalinkaaaa14 (changes requested):

shared/src/packages/drivers/libs/validation-schemas/driver-request-body.validation-schema.ts
kalinkaaaa14, line 1: you forgot to rename files

shared/src/packages/drivers/libs/validation-schemas/driver-parameters.validation-schema.ts
kalinkaaaa14, line 1: same about file renaming

shared/src/packages/drivers/libs/types/response/driver-create-update-response-dto.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type UserCommonDetails } from '~/packages/users/users.js';
+
+import { type DriverEntity } from '../driver-entity.type.js';
+
+type DriverCreateUpdateResponseDto = UserCommonDetails & {
+  driver: DriverEntity;
+};
```
kalinkaaaa14:
```suggestion
type DriverCreateUpdateResponseDto = UserCommonDetails & DriverEntity;
```

shared/src/packages/drivers/libs/types/service/driver-add-payload.type.ts
```diff
@@ -0,0 +1,10 @@
+import { type DriverDto } from '../driver-dto.type.js';
+
+type DriverAddPayload = {
+  payload: DriverDto & {
+    password: string;
+  };
```
kalinkaaaa14: you have already DriverCreateUpdateRequestDto type created

shared/src/packages/drivers/libs/types/service/driver-update-payload.ts
```diff
@@ -0,0 +1,10 @@
+import { type DriverDto } from '../types.js';
+
+type DriverUpdatePayload = {
+  payload: DriverDto & {
+    password: string;
+  };
```
kalinkaaaa14: you have already DriverCreateUpdateRequestDto type created

shared/src/packages/drivers/libs/types/request/driver-get-all-by-business-request-dto.type.ts
```diff
@@ -0,0 +1,3 @@
+type DriverAllByBusinessRequestParameters = { businessId: number };
```
kalinkaaaa14: this type is more appropriate to business package I think, and can be named like
type BusinessGetAllDriversRequestParameters

PavloSukhinin (dismissed):

Druxight (dismissed):

kalinkaaaa14 (approved):

DmytroVoronkov (approved):

Druxight (approved):

PR #148, opened 2023-09-01
th-126: The Password field content on the Sign up both Business and Customer and Sign in pages is hidden by asterisks after the Show Hide password button was clicked

albsbz:
Add show/hide password behaviour
pw1
pw2


frontend/src/libs/components/input/input.tsx
```diff
@@ -42,19 +46,29 @@ const Input = <T extends FieldValues>({
...
           {...field}
-          type={type}
+          type={showPassword ? 'text' : type}
           placeholder={placeholder}
           className={getValidClassNames(...inputStyles)}
           disabled={isDisabled}
         />
         {type === 'password' && (
-          <span className={styles.passwordEye}>&#128065;</span>
+          <input
```
kalinkaaaa14: can we add it as button, not as input type="button"? 
also what about changing eye icon when turning on/off? now it's not working this way

albsbz: fixed

frontend/src/libs/components/input/input.tsx
```diff
@@ -30,7 +36,7 @@ const Input = <T extends FieldValues>({
   isDisabled,
 }: Properties<T>): JSX.Element => {
   const { field } = useFormController({ name, control });
-
+  const [showPassword, setShowPassword] = useState(false);
```
h0wter, line 39: It is better to call boolean using be/have verbs. For example: `isPasswordShown`.

h0wter: It is better to call boolean using be/have verbs. For example: `isPasswordShown`.

frontend/src/libs/components/input/styles.module.scss
```diff
@@ -86,7 +86,10 @@ $error-message-height: calc($error-font-size * $line-height-coefficient);
 .passwordEye {
   position: absolute;
   top: 8px;
-  right: 15px;
+  right: 10px;
+  color: $grey-light;
+  background-color: $white;
```
h0wter, line 91: Maybe we should make different state icons have different colors?

albsbz: You mean on hover, or depending if password showed or not?

h0wter: I meant depending on the state, shown or not.

DmytroVoronkov (approved):

frontend/src/libs/components/input/styles.module.scss
albsbz: Added color changing

h0wter (approved):

frontend/src/libs/components/input/input.tsx
albsbz: fixed

frontend/src/libs/components/input/input.tsx
```diff
@@ -30,7 +36,7 @@ const Input = <T extends FieldValues>({
   isDisabled,
 }: Properties<T>): JSX.Element => {
   const { field } = useFormController({ name, control });
-
+  const [isShowPassword, setIsShowPassword] = useState(false);
```
kalinkaaaa14:
```suggestion
  const [isPasswordShown, setIsPasswordShown] = useState(false);
```

frontend/src/libs/components/input/input.tsx
```diff
@@ -42,19 +48,36 @@ const Input = <T extends FieldValues>({
...
         />
         {type === 'password' && (
-          <span className={styles.passwordEye}>&#128065;</span>
+          <button
+            className={getValidClassNames(
+              styles.passwordEye,
+              isShowPassword && styles.passwordEyeLight,
+            )}
+            onClick={toggleShowPassword}
+            tabIndex={-1}
```
kalinkaaaa14, line 86: do we need tabIndex with button here?

frontend/src/libs/components/input/input.tsx
```diff
@@ -42,19 +48,36 @@ const Input = <T extends FieldValues>({
     hasError && styles.error,
   ];
 
+  const toggleShowPassword = useCallback(
+    (event: React.MouseEvent<HTMLElement>): void => {
+      event.preventDefault();
```
kalinkaaaa14, line 59: can we remove prevent default? for which purpose it's used here

frontend/src/libs/components/input/input.tsx
albsbz: If remove it when tabing through form eye symbol will be in focus,  It seems for me as not predictable behavior

frontend/src/libs/components/input/input.tsx
albsbz: When I remove it click event bubbles up and triggers form as touched, so all field becomes red

frontend/src/libs/components/input/input.tsx
albsbz: fixed

kalinkaaaa14 (dismissed):

frontend/src/libs/components/input/input.tsx
```diff
@@ -5,9 +5,15 @@ import {
   type FieldValues,
 } from 'react-hook-form';
 
+import { IconName } from '~/libs/enums/icon-name.enum';
```
PavloSukhinin: add file extension, please

PavloSukhinin (changes requested):

PavloSukhinin (approved):

kalinkaaaa14 (approved):

Druxight (approved):

PR #150, opened 2023-09-01
th-110: Order schema

albsbz:
Added:
1. Orders scheme
2. Orders repository 
3. Orders controller 
4. Orders service
pako_eNrFV8lu2zAQ_RWCJxdwAu-xdS56aVEU6K0wINDS2BZCkSqXOErgf-9Qkh1ZiyUHDaKLYfLpzfJmhtQrDWQI1KOgvkZsp1i8FmtB8LEalCav-R_3RCHBpYhx8us7WVMhDRGW8zV9gyR7KYA8MRXsmSphhsSK6K-FMhZiFvGe2G2ktPEFi5vIy0DO-uESpvVBqtDfM73vi9WMmw7sTkmb


backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -60,4 +63,51 @@ const business = pgTable('business_details', {
...
+const businessRelations = relations(users, ({ many }) => ({
+  orders: many(orders),
+}));
+
+const orderStatuses = Object.values(OrderStatus as Record<number, string>) as [
+  string,
+  ...string[],
+];
+
+const statusEnum = pgEnum('status_enum', orderStatuses);
```
PavloSukhinin: let's create helper for this and keep our schemas file clean

backend/src/packages/orders/order.controller.ts
```diff
@@ -0,0 +1,148 @@
...
+import { type OrderService } from '~/packages/orders/order.service.js';
+
+import { OrdersApiPath } from './libs/enums/enums.js';
+import {
+  type OrderCreateRequestDto,
+  type OrderEntity,
+  type OrderUpdateRequestDto,
+} from './libs/types/types.js';
+
+type idT = Pick<OrderEntity, 'id'>;
```
PavloSukhinin: - type names must be capitalized
- can we have shared type for ID?

shared/src/packages/orders/libs/types/order-status.type.ts
```diff
@@ -0,0 +1,5 @@
+import { type OrderStatus } from '../enums/enums.js';
+
+type OrderStatus = (typeof OrderStatus)[keyof typeof OrderStatus];
```
PavloSukhinin: Just use ValueOf<typeof OrderStatus> when needed

shared/src/packages/orders/libs/types/order-entity.type.ts
```diff
@@ -0,0 +1,17 @@
...
+  price: number;
+  scheduledTime: string;
+  startPoint: string;
+  endPoint: string;
+  status: OrderStatus;
+  userId: number;
+  businessId: number;
+  driverId: number;
+  customerName?: string;
+  customerPhone?: string;
```
PavloSukhinin:
```suggestion
  driverId: number;
  customerName: string | null;
  customerPhone: string | null;
```
🤔

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  public constructor(orderRepository: OrderRepository) {
+    this.orderRepository = orderRepository;
+  }
+
+  public async create(
+    payload: OrderCreateRequestDto,
+    userId = 1, //Mock, get from JWT
+  ): Promise<OrderCreateResponseDto> {
+    const price = 100; //Mock
```
PavloSukhinin: we already have jwt, let's remove mock?

backend/src/packages/orders/order.entity.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  private userId: OrderEntityT['userId'];
+
+  private businessId: OrderEntityT['businessId'];
+
+  private driverId: OrderEntityT['driverId'];
+
+  private customerName: OrderEntityT['customerName'] | undefined;
+
+  private customerPhone: OrderEntityT['customerPhone'] | undefined;
```
PavloSukhinin:
```suggestion
  private customerName: OrderEntityT['customerName'];

  private customerPhone: OrderEntityT['customerPhone'];
```
In another comment I mentioned that customerName should be defined, but nullable `| null`

shared/src/packages/orders/libs/types/order-entity.type.ts
```diff
@@ -0,0 +1,17 @@
+import { type OrderStatus } from './order-status.type.js';
+
+type OrderEntity = {
+  id: number;
+  price: number;
+  scheduledTime: string;
+  startPoint: string;
+  endPoint: string;
+  status: OrderStatus;
+  userId: number;
```
PavloSukhinin: it's nullable, because we'll allow to make orders without registration

backend/src/packages/orders/order.controller.ts
```diff
@@ -0,0 +1,148 @@
...
+
+class OrderController extends Controller {
```
PavloSukhinin, line 199: Let's add swagger doc?

PavloSukhinin (changes requested):

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,140 @@
...
+    //
+  }
+
+  public delete(): Promise<boolean> {
+    return Promise.resolve(false);
+  }
+
+  public create(): Promise<unknown> {
+    return Promise.resolve();
+  }
```
PavloSukhinin: `class OrderService implements Omit<IService, 'find' | 'create'>`
and then remove these methods. 
let's discuss it with the team

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (!foundOrder) {
+      throw new HttpError({
+        status: HttpCode.NOT_FOUND,
+        message: HttpMessage.ORDER_DOES_NOT_EXIST,
+      });
+    }
+
+    const orderBelongsToUser =
+      (foundOrder.userId && foundOrder.userId === userId) ||
+      (foundOrder.customerPhone && foundOrder.customerPhone === phone);
```
PavloSukhinin, line 239: `foundOrder.userId && foundOrder.userId === userId` I get it, we can create order without userId. 
But not sure that same can be applied to phone, can we just check it like this `foundOrder?.customerPhone === phone`?

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,140 @@
...
+      items: usersOrders.map((it) => it.toObject()),
+    };
+  }
+
+  public async deleteIfBelongsToUser(
+    id: OrderEntityT['id'],
+    user: UserEntityObjectWithGroupT,
+  ): Promise<boolean> {
+    const { id: userId, phone } = user;
+    const { result: foundOrder } = await this.findById(id);
```
PavloSukhinin: I saw we removed OperationResult type in one of the previous PR's. Discuss this the team - what should be returned from repository.

shared/src/packages/orders/libs/enums/orders-validation-message.enum.ts
```diff
@@ -0,0 +1,9 @@
+const OrdersValidationMessage = {
+  // COMPANY_NAME_REQUIRED: 'Company name is required',
+  // TAX_NUMBER_REQUIRED: 'Tax number is required',
```
PavloSukhinin: Will we use it later? If not - then remove commented code

PavloSukhinin (changes requested):

backend/src/packages/orders/order.service.ts
albsbz: As I see it if we create order with user, user already has phone, so we don't duplicate this in order table. In such case order can be with userId either with customerPhone

backend/src/libs/packages/database/schema/tables-schema.ts
albsbz: done

backend/src/packages/orders/order.controller.ts
albsbz: created shared type for Id

shared/src/packages/orders/libs/types/order-status.type.ts
albsbz: Don't noticed it before, used ValueOf as suggested

shared/src/packages/orders/libs/types/order-entity.type.ts
albsbz: fixed

backend/src/packages/orders/order.service.ts
albsbz: removed JWT mock

backend/src/packages/orders/order.entity.ts
albsbz: fixed

shared/src/packages/orders/libs/types/order-entity.type.ts
albsbz: got it, fixed

backend/src/packages/orders/order.controller.ts
albsbz: added

backend/src/packages/orders/order.service.ts
albsbz: fixed, made create method accept userId as part of payload parameter

shared/src/packages/orders/libs/enums/orders-validation-message.enum.ts
albsbz: removed

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,123 @@
...
+    database: Pick<IDatabase, 'driver'>,
+    ordersSchema: DatabaseSchema['orders'],
+  ) {
+    this.db = database;
+    this.ordersSchema = ordersSchema;
+  }
+
+  public find(): Promise<InferModel<typeof schema.users>[]> {
+    return Promise.resolve([]);
+  }
```
DmytroVoronkov: What does this code do?

albsbz: missed it, I'll remove it. will use `class OrderRepository implements Omit<IRepository, 'find'>` instead

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
```diff
@@ -0,0 +1,19 @@
+import joi from 'joi';
+
+import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enums/user-validation-rule.enum.js';
```
kalinkaaaa14:
```suggestion
import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enums/enums.js';
```

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { OrderStatus } from '../enums/order-status.enum.js';
+import { type OrderUpdateRequestDto } from '../types/order-update-request-dto.type.js';
+
+const orderUpdate = joi.object<OrderUpdateRequestDto, true>({
```
kalinkaaaa14:
```suggestion
const orderUpdateRequestBody = joi.object<OrderUpdateRequestDto, true>({
```

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
```diff
@@ -0,0 +1,19 @@
+import joi from 'joi';
+
+import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enums/user-validation-rule.enum.js';
+
+import { OrderStatus } from '../enums/order-status.enum.js';
```
kalinkaaaa14:
```suggestion
import { OrderStatus } from '../enums/enums.js';
```

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
```diff
@@ -0,0 +1,19 @@
...
+import { type OrderUpdateRequestDto } from '../types/order-update-request-dto.type.js';
```
kalinkaaaa14:
```suggestion
import { type OrderUpdateRequestDto } from '../types/types.js';
```

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const orderUpdate = joi.object<OrderUpdateRequestDto, true>({
+  scheduledTime: joi.string().isoDate(),
+  startPoint: joi.string(),
+  endPoint: joi.string(),
+  driverId: joi.number().integer().positive(),
+  userId: joi.number().integer().positive(),
+  status: joi.string().valid(...Object.values(OrderStatus)),
+  customerName: joi.string().pattern(UserValidationRule.NAME),
+  customerPhone: joi.string().pattern(UserValidationRule.PHONE),
```
kalinkaaaa14: maybe not all of these fields can be updated in order ? for example userId/driverId/startPoint/endPoint ? Im not really sure that this data can be updated when the order is created already

shared/src/packages/orders/libs/validation-schemas/orders-find-by.validation-schema.ts
```diff
@@ -0,0 +1,18 @@
+import joi from 'joi';
+
+import { type OrderFindByRequestDto } from '../types/order-find-by-request-dto.type.js';
+
+const orderFindBy = joi.object<OrderFindByRequestDto, true>({
```
kalinkaaaa14:
```suggestion
const orderGetQueryParams = joi.object<OrderFindByRequestDto, true>({
```

shared/src/packages/orders/libs/validation-schemas/orders-create.validation-schema.ts
```diff
@@ -0,0 +1,16 @@
+import joi from 'joi';
+
+import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enums/user-validation-rule.enum.js';
+
+import { type OrderCreateRequestDto } from '../types/order-create-request-dto.type.js';
+
+const orderCreate = joi.object<OrderCreateRequestDto, true>({
```
kalinkaaaa14:
```suggestion
const orderCreateRequestBody = joi.object<OrderCreateRequestDto, true>({
```

shared/src/packages/orders/libs/validation-schemas/order-id-parameter.validation-schema.ts
```diff
@@ -0,0 +1,13 @@
+import joi from 'joi';
+
+import { type Id } from '~/libs/types/route-parameters.type.js';
+
+import { OrdersValidationMessage } from '../enums/enums.js';
+
+const orderIdParameter = joi.object<Id, true>({
```
kalinkaaaa14:
```suggestion
const orderGetParameter = joi.object<Id, true>({
```

shared/src/packages/orders/libs/types/order-status.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type ValueOf } from '~/libs/types/value-of.type.js';
+
+import { type OrderStatus } from '../enums/enums.js';
```
kalinkaaaa14: why with type ?
```suggestion
import { OrderStatus } from '../enums/enums.js';
```

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+import { type OrderEntity } from './order-entity.type.js';
+
+type OrderFindByRequestDto = Partial<Omit<OrderEntity, 'id'>>;
```
kalinkaaaa14: won't it be suitable?
```suggestion
type OrderFindByRequestDto = Omit<OrderEntity, 'id'>;
```

backend/src/packages/orders/orders.ts
```diff
@@ -0,0 +1,17 @@
+import { database, schema } from '~/libs/packages/database/database.js';
+import { logger } from '~/libs/packages/logger/logger.js';
+
+import { businessService } from '../business/business.js';
```
kalinkaaaa14, line 4: why not BusinessService ?

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,181 @@
...
+class OrderService implements Omit<IService, 'delete' | 'findById'> {
+  private orderRepository: OrderRepository;
+
+  private errors: Record<
+    string,
+    {
+      status: ValueOf<typeof HttpCode>;
+      message: string;
+    }
+  >;
```
kalinkaaaa14: Im not sure we need this private field, we can just throw errors if something goes wrong

kalinkaaaa14 (changes requested):

backend/src/libs/packages/database/schema/database-enums/order-status.ts
```diff
@@ -0,0 +1,11 @@
...
+import { OrderStatus } from 'shared/build/index.js';
+
+const orderStatuses = Object.values(OrderStatus as Record<number, string>) as [
+  string,
+  ...string[],
+];
+
+const statusEnum = pgEnum('status_enum', orderStatuses);
+
+export { statusEnum };
```
kalinkaaaa14:
```
const ORDER_STATUSES = Object.values(OrderStatus);

const STATUS_VALUES = pgEnum('status', ORDER_STATUSES);
```

kalinkaaaa14 (changes requested):

backend/src/packages/orders/order.service.ts
albsbz: removed OperationResult usage

backend/src/packages/orders/orders.ts
albsbz: We make here DI of instance of business service into business controller, as its instance I use lower case

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
albsbz: I tried to make universal method that can update different fields, and some can be present, some not if at some point we know that we need only some field it can be more stricted

shared/src/packages/orders/libs/types/order-status.type.ts
albsbz: there is arule for eslint that not allows to import without type: All imports in the declaration are only used as types. Use `import type`.eslint[@typescript-eslint/consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports)

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
albsbz: I agree, I 'm realy not sure about all of them except status, I think if we decided to correct route, or maybe swap driver for the same price, something like that than we can left it.

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
albsbz: fixed

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
albsbz: fixed

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
albsbz: fixed

shared/src/packages/orders/libs/validation-schemas/orders-update.validation-schema.ts
albsbz: fixed

shared/src/packages/orders/libs/validation-schemas/orders-find-by.validation-schema.ts
albsbz: fix

shared/src/packages/orders/libs/validation-schemas/orders-create.validation-schema.ts
albsbz: fix

shared/src/packages/orders/libs/validation-schemas/order-id-parameter.validation-schema.ts
albsbz: fixed

backend/src/packages/orders/order.service.ts
albsbz: replace with custom error for orders

backend/src/libs/packages/database/schema/database-enums/order-status.ts
albsbz: renamed

backend/src/libs/packages/database/generated-schema/meta/_journal.json
```diff
@@ -36,6 +36,20 @@
...
+      "tag": "0005_add_orders_table",
+      "breakpoints": true
+    },
+    {
+      "idx": 6,
+      "version": "5",
+      "when": 1693914703726,
+      "tag": "0005_add_orders_table",
+      "breakpoints": true
     }
```
PavloSukhinin: something wrong with this meta

backend/src/packages/orders/order.service.ts
PavloSukhinin: Got it, thanks 👍

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,115 @@
...
+    payload: Partial<OrderEntityT>;
+  }): Promise<OrderEntity | null> {
+    const result = await this.db
+      .driver()
+      .update(this.ordersSchema)
+      .set(payload)
+      .where(eq(this.ordersSchema.id, id))
+      .returning();
+
+    const order = result[0] as OrderEntityT;
```
PavloSukhinin: `const [order = null] = await...`
I'd prefer destructuring

backend/src/libs/packages/database/generated-schema/0005_add_orders_table.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE IF NOT EXISTS
+  "orders" (
+    "id" serial PRIMARY KEY NOT NULL,
+    "price" integer NOT NULL,
+    "scheduled_time" timestamp NOT NULL,
+    "start_point" varchar NOT NULL,
+    "end_point" varchar NOT NULL,
+    "status" "status" NOT NULL,
```
PavloSukhinin: is this a valid syntax?

backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -1,12 +1,14 @@
...
 import {
   integer,
+  pgEnum,
   pgTable,
   serial,
   timestamp,
   uniqueIndex,
   varchar,
 } from 'drizzle-orm/pg-core';
+import { ORDER_STATUSES } from 'shared/build/index.js';
```
PavloSukhinin, line 13: Ok, this is the only place where we'll allow it (direct import from shared)

PavloSukhinin (changes requested):

backend/src/libs/packages/database/schema/tables-schema.ts
albsbz: got it

backend/src/packages/orders/order.repository.ts
albsbz: cool, didn't know about destructuring assignment

backend/src/libs/packages/database/generated-schema/0005_add_orders_table.sql
albsbz: yes, but I forgot to export Enum from tables-schema.ts, and it was reason why `CREATE TYPE "status" AS ENUM('pending', 'confirmed', 'canceled', 'done');` was not generated. Fixed

backend/src/libs/packages/database/generated-schema/meta/_journal.json
albsbz: fixed

backend/src/libs/packages/database/schema/tables-schema.ts
```diff
@@ -60,4 +63,45 @@ const business = pgTable('business_details', {
   updatedAt: timestamp('updated_at').notNull().defaultNow(),
 });
 
-export { business, groups, users, usersRelations };
+const businessRelations = relations(users, ({ many }) => ({
+  orders: many(orders),
+}));
+
+const orderStatus = pgEnum('status', ORDER_STATUSES);
```
PavloSukhinin:
```suggestion
const orderStatus = pgEnum('order_status', ORDER_STATUSES);
```
to avoid collisions in the future - there might be many different statuses

DmytroVoronkov (approved):

backend/src/libs/packages/database/schema/tables-schema.ts
albsbz: fixed

shared/src/packages/orders/libs/validation-schemas/order-get-parameter.validation-schema.ts
```diff
@@ -0,0 +1,13 @@
+import joi from 'joi';
+
+import { type Id } from '~/libs/types/types.js';
+
+import { OrdersValidationMessage } from '../enums/enums.js';
+
+const orderGetParameter = joi.object<Id, true>({
+  id: joi.number().integer().positive().required().messages({
```
MaxHopaytsa: There is a scheme in development to validate a positive id number, you can use it. It is located here: shared\src\libs\validation-schemas\validation-schemas.ts

albsbz: cool, fixed

shared/src/packages/orders/libs/validation-schemas/order-get-query-parameters.validation-schema.ts
```diff
@@ -0,0 +1,11 @@
+import joi from 'joi';
+
+import { type OrderFindByRequestDto } from '../types/types.js';
+
+const orderGetQueryParameters = joi.object<OrderFindByRequestDto, true>({
+  driverId: joi.string(),
+  userId: joi.string(),
+  businessId: joi.string(),
```
kalinkaaaa14: why strings?

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
albsbz: fixed only to driverId, userId and businessId

shared/src/packages/orders/libs/validation-schemas/order-create-request-body.validation-schema.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { positiveRequiredIntegerSchema } from '~/libs/validation-schemas/validation-schemas.js';
+import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enums/enums.js';
+
+import { OrdersValidationMessage } from '../enums/orders-validation-message.enum.js';
+import { type OrderCreateRequestDto } from '../types/types.js';
+
+const orderCreateRequestBody = joi.object<OrderCreateRequestDto, true>({
+  scheduledTime: joi.string().isoDate().required(),
+  carsQty: joi.number().integer().positive().required(),
```
kalinkaaaa14: can you reuse positiveRequiredInteregerSchema there? do we have some requirements about min/max numbers?

shared/src/packages/orders/libs/types/order-status.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type ValueOf } from '~/libs/types/value-of.type.js';
+
+import { type OrderStatus } from '../enums/order-status.enum.js';
+
+type OrderStatus = ValueOf<typeof OrderStatus>;
```
kalinkaaaa14: it's better to rename it, for ex type OrderStatusValues

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type OrderFindByRequestDto = {
+  'userId': string;
+  'businessId': string;
+  'driverId': string;
+};
```
kalinkaaaa14: why props in single quotes? why strings?

shared/src/packages/orders/libs/constants/order-statuses.ts
```diff
@@ -0,0 +1,8 @@
+import { OrderStatus } from '../enums/order-status.enum.js';
+
+const ORDER_STATUSES = Object.values(OrderStatus as Record<number, string>) as [
+  string,
+  ...string[],
+];
+
```
kalinkaaaa14: In my last comment I proposed like this, why don't you change?
```suggestion
import { OrderStatus } from '../enums/order-status.enum.js';

const ORDER_STATUSES = Object.values(OrderStatus);

```

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,168 @@
...
+    const foundOrder = await this.findById(id, user);
+
+    if (!foundOrder) {
+      throw new OrderNotFound();
+    }
+
+    return await this.orderRepository.delete(id);
+  }
+
+  private checkIfOrderBelongsToDriver(
```
kalinkaaaa14: methods that are called with check are usually retuning boolean value. So, you should better rename method, or return boolean value from this method

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,168 @@
...
+    businessId,
+    driverId,
+    currentUserBusinessId,
+    currentUserId,
+  }: {
+    userId: string | undefined;
+    businessId: string | undefined;
+    driverId: string | undefined;
+    currentUserBusinessId: number | undefined;
+    currentUserId: number | null;
```
kalinkaaaa14: if you want to write such generic find then use ?
and again, why ids are strings ?
for me, I don't like this idea of almost all undefined fields. I prefer more approach with separate methods for finding, but each for special purpose.

```suggestion
    userId?: string;
    businessId?: string;
    driverId?: string;
    currentUserBusinessId:? number | undefined;
    currentUserId: number | null;
```

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,111 @@
...
+      carsQty,
+      startPoint,
+      endPoint,
+      status,
+      userId,
+      businessId,
+      driverId,
+      customerName,
+      customerPhone,
+    } = entity.toNewObject();
```
kalinkaaaa14: do you need to destruct each prop?

backend/src/packages/orders/libs/exceptions/order-not-found.ts
```diff
@@ -0,0 +1,12 @@
+import { HttpCode, HttpError, HttpMessage } from 'shared/build/index.js';
+
+class OrderNotFound extends HttpError {
```
kalinkaaaa14: we have already NotFoundError

kalinkaaaa14 (changes requested):

shared/src/packages/orders/libs/validation-schemas/order-get-query-parameters.validation-schema.ts
albsbz: I make them strings because it comes from query, but I'll check if joy will interprit them as number

shared/src/packages/orders/libs/constants/order-statuses.ts
albsbz: When I tried to do it in this way in line `const orderStatus = pgEnum('order_status', ORDER_STATUSES);` I get 
```
Argument of type '("pending" | "confirmed" | "canceled" | "done")[]' is not assignable to parameter of type 'readonly [string, ...string[]] | [string, ...string[]]'.
  Type '("pending" | "confirmed" | "canceled" | "done")[]' is not assignable to type '[string, ...string[]]'.
```

backend/src/packages/orders/order.service.ts
albsbz: ok, I'll use '?', also I'll change type to 'number', as joy worked with query strings interpreting string like numbers, and using string type was not good idea. 
I suppose that at some point we will use filters to find all orders made by driver that belongs to business, all orders  made by user from particular business and so on. 
I can suggest to make this method private and use it from other methods. 
What do you think? But If you insist I l'll make separate methods

backend/src/packages/orders/libs/exceptions/order-not-found.ts
albsbz: Yes, we have NotFoundError which extends ApplicationError, not HttpError and when throw it returns 500 status, as I throw error with same messsage in several places in order package I added custom error that extends HttpError just for not duplicating code

backend/src/packages/orders/order.repository.ts
albsbz: surely not, removed it

backend/src/packages/orders/order.service.ts
albsbz: fixed

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
albsbz: fixed it, it was a bad idea)

shared/src/packages/orders/libs/types/order-status.type.ts
albsbz: done

shared/src/packages/orders/libs/validation-schemas/order-create-request-body.validation-schema.ts
albsbz: done

shared/src/packages/orders/libs/validation-schemas/order-get-query-parameters.validation-schema.ts
albsbz: fixed

backend/src/packages/orders/libs/exceptions/order-not-found.ts
kalinkaaaa14: you should try again reuse NotFoundError, it was used for business and it's basic purpose was to create general error for such cases

backend/src/packages/orders/order.service.ts
Some14u: I'm also more towards having separate findById, findByUserId, findBySomething inside services.
Although, I'd consider to have a more versatile "findByAnything"-like method as part of your orderRepository and use it in every findBy... of your service.

backend/src/packages/auth/auth.controller.ts
```diff
@@ -30,7 +30,7 @@ class AuthController extends Controller {
     this.authService = authService;
 
     this.addRoute({
-      path: AuthApiPath.SIGN_UP_CUSTOMER,
+      path: AuthApiPath.SIGN_UP,
       method: 'POST',
       validation: {
```
TetianaDoroshko: Doesn't it ruin a sign-up logic?

albsbz: removed it, it was mock from when signup didn't work in another way

backend/src/packages/orders/order.controller.ts
```diff
@@ -0,0 +1,616 @@
...
+      },
+      handler: (options) =>
+        this.updateOne(
+          options as ApiHandlerOptions<{
+            params: Id;
+            body: OrderUpdateRequestDto;
+            user: UserEntityObjectWithGroupT;
+          }>,
+        ),
+    });
```
kalinkaaaa14: why post for update

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+import { type OrderEntity } from './order-entity.type.js';
+
+type OrderFindByRequestDto = Pick<
```
kalinkaaaa14: do we still use it somewhere in the code?

backend/src/packages/orders/libs/exceptions/order-not-found.ts
```diff
@@ -0,0 +1,12 @@
+import { HttpMessage } from '~/libs/enums/enums.js';
+import { NotFoundError } from '~/libs/exceptions/exceptions.js';
+
+class OrderNotFound extends NotFoundError {
+  public constructor() {
+    super({
+      message: HttpMessage.ORDER_DOES_NOT_EXIST,
+    });
+  }
+}
```
kalinkaaaa14: again, can't you reuse not found error and throw there needed message? now it looks like same error but with the other message

backend/src/packages/orders/libs/helpers/combine-filters.ts
```diff
@@ -0,0 +1,21 @@
...
+import { type PgColumn } from 'drizzle-orm/pg-core';
+
+import { type DatabaseSchema } from '~/libs/packages/database/schema/schema.js';
+
+type Table = DatabaseSchema[keyof DatabaseSchema];
+
+const combineFilters = <TSchema extends Table>(
+  schema: TSchema,
+  searchBy: Record<PgColumn['name'], string | number | null>,
+): SQL | undefined => {
```
kalinkaaaa14, line 11: why undefined can be returned?

backend/src/packages/orders/order.controller.ts
```diff
@@ -0,0 +1,616 @@
...
+    this.orderService = orderService;
+
+    this.businessService = businessService;
+
+    this.driverService = driverService;
+
+    this.addRoute({
+      path: OrdersApiPath.ROOT,
+      method: 'GET',
+      authStrategy: AuthStrategy.VERIFY_JWT,
```
kalinkaaaa14: this route should be available only for business, is it now implemented such way?

backend/src/packages/orders/libs/helpers/combine-filters.ts
albsbz: It's can be undefined beacause drizzle-orm `and()` operator potentially can return undefined type if conditions not passed

shared/src/packages/orders/libs/types/order-find-by-request-dto.type.ts
albsbz: removed

backend/src/packages/orders/order.controller.ts
albsbz: change to put

backend/src/packages/orders/order.controller.ts
albsbz: restricted to business

backend/src/packages/orders/libs/exceptions/order-not-found.ts
albsbz: done

shared/src/packages/orders/libs/types/order-with-driver-entity.type.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { type OrderEntity } from './order-entity.type.js';
+
+type OrderWithDriverEntity = OrderEntity & {
+  driver: Pick<DriverEntity, 'driverLicenseNumber'> & {
+    user: {
+      firstName: UserEntityT['firstName'];
+      lastName: UserEntityT['lastName'];
+    };
+  };
```
kalinkaaaa14: can u reuse DriverDto here?

backend/src/packages/orders/libs/exceptions/business-not-exist.ts
```diff
@@ -0,0 +1,12 @@
+import { HttpCode, HttpError, HttpMessage } from 'shared/build/index.js';
+
+class BusinessNotExist extends HttpError {
```
kalinkaaaa14: can't you reuse not found error and throw there needed message?

backend/src/packages/orders/libs/exceptions/driver-not-exist.ts
```diff
@@ -0,0 +1,12 @@
+import { HttpCode, HttpError, HttpMessage } from 'shared/build/index.js';
+
+class DriverNotExist extends HttpError {
```
kalinkaaaa14: can't you reuse not found error and throw there needed message? 
I don't like that you are duplicating code for each entity

backend/src/packages/orders/order-with-driver.entity.ts
```diff
@@ -0,0 +1,88 @@
+import { type IEntity } from '~/libs/interfaces/interfaces.js';
+import { type NullableProperties } from '~/libs/types/types.js';
+
+import { type OrderWithDriverEntity as OrderWithDriverEntityT } from './libs/types/types.js';
+
+class OrderWithDriverEntity implements Omit<IEntity, 'toNewObject'> {
```
kalinkaaaa14: why do you need this entity, it's duplicating orderEntity

kalinkaaaa14: I think you don't need it in entities, remove

kalinkaaaa14 (changes requested):

shared/src/packages/orders/libs/types/order-find-by-id-response-dto.type.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { type OrderEntity } from './order-entity.type.js';
+
+type OrderFindByIdResponseDto = OrderEntity & {
+  driver: Pick<DriverDto, 'driverLicenseNumber'> & {
+    user: {
+      firstName: UserEntityT['firstName'];
+      lastName: UserEntityT['lastName'];
+    };
+  };
```
kalinkaaaa14: I meant to do like this, to avoid such structure driver: { user: fields } }
```suggestion
  driver: DriverDto;
  };
```

backend/src/packages/orders/order.controller.ts
```diff
@@ -0,0 +1,616 @@
...
+    options: ApiHandlerOptions<{
+      params: Id;
+      user: UserEntityObjectWithGroupT | null;
+    }>,
+  ): Promise<ApiHandlerResponse> {
+    let business;
+
+    if (options.user?.group.key === UserGroupKey.BUSINESS) {
+      business = await this.businessService.findByOwnerId(options.user.id);
+    }
```
kalinkaaaa14: you don't need this block in controller layer. and you don't need to pass businessId, just pass user, and then in service inside this block  find business id and verify if order belongs to this business
  if (user?.group.key === UserGroupKey.BUSINESS) {}

kalinkaaaa14 (changes requested):

backend/src/libs/exceptions/not-found-error/not-found-error.exception.ts
```diff
@@ -1,23 +1,18 @@
...
   cause?: unknown;
 };
 
-class NotFoundError extends ApplicationError {
-  public status: ValueOf<typeof HttpCode>;
-
+class NotFoundError extends HttpError {
   public constructor({ message, cause }: Constructor) {
     super({
       message: message ?? AppErrorMessage.ENTITY_NOT_FOUND,
```
Some14u, line 12: I believe we should abandon AppErrorMessage (delete it entirely) and use shared\src\libs\packages\http\libs\enums\http-error-message.enum.ts

There are all serverside error messages, and I don't see the point in separating them like that.

backend/src/packages/orders/libs/helpers/combine-filters.ts
```diff
@@ -0,0 +1,21 @@
+import { type SQL, and, eq } from 'drizzle-orm';
+import { type PgColumn } from 'drizzle-orm/pg-core';
+
+import { type DatabaseSchema } from '~/libs/packages/database/schema/schema.js';
+
+type Table = DatabaseSchema[keyof DatabaseSchema];
+
+const combineFilters = <TSchema extends Table>(
```
Some14u, line 8: If we could reuse it in other schemas, then i would suggest to move it outside "orders" scope to some more generic folder inside backend package.

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,125 @@
...
+  ): Promise<OrderEntity[]> {
+    const orders = await this.db
+      .driver()
+      .select()
+      .from(this.ordersSchema)
+      .where(
+        combineFilters<DatabaseSchema['orders']>(this.ordersSchema, search),
+      );
+
+    return orders.map((it) => OrderEntity.initialize(it as OrderEntityT));
```
Some14u, line 81: I thought we agreed, that we create entity at Service layer

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,125 @@
...
+    payload: Partial<OrderEntityT>;
+  }): Promise<OrderEntity | null> {
+    const [result = null] = await this.db
+      .driver()
+      .update(this.ordersSchema)
+      .set(payload)
+      .where(eq(this.ordersSchema.id, id))
+      .returning();
+
+    return result && OrderEntity.initialize(result as OrderEntityT);
```
Some14u, line 109: here too

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  public async create(entity: OrderEntity): Promise<OrderEntity> {
+    const [result] = await this.db
+      .driver()
+      .insert(this.ordersSchema)
+      .values(entity.toNewObject())
+      .returning()
+      .execute();
+
+    return OrderEntity.initialize(result as OrderEntityT);
```
Some14u, line 92: and here

backend/src/packages/orders/order.repository.ts
```diff
@@ -0,0 +1,125 @@
...
+      driver: { ...modifiedDriver, ...user },
+    };
+  }
+
+  public async findById(id: OrderEntityT['id']): Promise<OrderEntity | null> {
+    const result = await this.db.driver().query.orders.findFirst({
+      where: (orders) => eq(orders.id, id),
+    });
+
+    return (result ?? null) && OrderEntity.initialize(result as OrderEntityT);
```
Some14u, line 67: see below

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,249 @@
...
+  public async findOne({
+    id,
+    user,
+  }: {
+    id: OrderEntityT['id'];
+    user: UserEntityObjectWithGroupT | null;
+    businessId?: number;
+  }): Promise<OrderFindByIdResponseDto | null> {
+    let businessId;
+
```
kalinkaaaa14: you don't need to create let variable which is not initailized

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,249 @@
...
+    return updatedOrder.toObject();
+  }
+
+  public async findOne({
+    id,
+    user,
+  }: {
+    id: OrderEntityT['id'];
+    user: UserEntityObjectWithGroupT | null;
+    businessId?: number;
```
kalinkaaaa14: it will be not passed here

kalinkaaaa14 (changes requested):

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,249 @@
...
+    if (!updatedOrder) {
+      throw new NotFoundError({
+        message: HttpMessage.ORDER_DOES_NOT_EXIST,
+      });
+    }
+
+    return updatedOrder.toObject();
+  }
+
+  public async findOne({
```
kalinkaaaa14, line 109:
```suggestion
  const result = await this.orderRepository.findByIdWithDriver(id);
  if (!result) {
    throw new NotFoundError({message: HttpMessage.ORDER_DOES_NOT_EXIST,});
  }

  const { order, driver } = result;
  //do we need create foundOrder var, or can we reuse result found from the repo
  const foundOrder = { ...order.toObject(), driver };

  if (user?.group.key === UserGroupKey.BUSINESS) {
    const business = await this.businessService.findByOwnerId(user.id);
    this.verifyOrderBelongsToBusiness(foundOrder, business);
}
  if (user) {
    this.verifyOrderBelongsToUser(foundOrder, user);
  }

  // how do we verify that user that is not logged can view this order?
  return foundOrder;
```

backend/src/packages/orders/order.service.ts
```diff
@@ -0,0 +1,249 @@
...
+        driverId,
+        customerName,
+        customerPhone,
+      }),
+    );
+
+    return order.toObject();
+  }
+
+  public async findById(id: OrderEntityT['id']): Promise<OrderEntityT | null> {
```
Some14u, line 82: How could it return null?
Check the same below...

backend/src/packages/orders/order.service.ts
albsbz: //do we need create foundOrder var, or can we reuse result found from the repo
No we can't reuse result from repo because entity of order not include driver, and we can't add driver to order entity because when we create order we should not add driver to it, So we can't use `order.toObject()` to entity returned from repo
 // how do we verify that user that is not logged can view this order?
we can't verify it for now, because we need some kind of token to identify unregistered user

backend/src/packages/orders/order.service.ts
albsbz: I need it because I set value inside one if statement and use it in another if statement, if I not initialize it in outer scope it will not be visible in scope of second if statement

kalinkaaaa14: I left an example below how you can rewrite method in service

shared/src/packages/orders/libs/constants/order-statuses.ts
```diff
@@ -0,0 +1,8 @@
+import { OrderStatus } from '../enums/order-status.enum.js';
+
+const ORDER_STATUSES = Object.values(OrderStatus as Record<number, string>) as [
+  string,
+  ...string[],
+];
```
Some14u:
```suggestion
const ORDER_STATUSES = Object.values(OrderStatus) as [
  ValueOf<typeof OrderStatus>,
];
```

This should keep it typed yet compatible with pgEnum argument

Some14u (changes requested):

backend/src/packages/orders/order.repository.ts
albsbz: It's seems like we use this approach in drivers, business and trucks repositories, if we decide to change type of database in future it allows as to encupsulate logic of converting Db entity to App entity inside repository layer

backend/src/packages/orders/order.service.ts
albsbz: answeared  below

backend/src/packages/orders/order.repository.ts
albsbz: drizzle findFirst() returns result | null

backend/src/packages/orders/order.repository.ts
albsbz: here we dont return null

backend/src/packages/orders/order.repository.ts
albsbz: see line ` const [result = null] = await this.db....`, if I understand you right

backend/src/packages/orders/libs/helpers/combine-filters.ts
albsbz: I'm not sure about it for now because it's not fully customizable because in line `searchBy: Record<PgColumn['name'], number | null>,`  where `number | null` we should list all posible types of PgColums. May be if refactoring, we can think about it

backend/src/libs/exceptions/not-found-error/not-found-error.exception.ts
albsbz: AppErrorMessage maybe used only in one place in some middelware where we catch all unhandled errors to not crash the App

PavloSukhinin (dismissed):

Druxight (dismissed):

kalinkaaaa14 (approved):

Druxight (approved):

PR #156, opened 2023-09-03
th-125: Fix content positioning below the header

h0wter:
- Added the ability to put routes into `PageLayout` and display them via `RouterOutlet`.
- Wrapped all the pages in `PageLayout`.
- Fixed the styles of auth pages so that they would display correctly inside `PageLayout` and made padding according to the design.
- Removed the ability to highlight password hide/display icons.


AlexVlaso (approved):

DmytroVoronkov (approved):

kalinkaaaa14 (dismissed):

frontend/src/libs/components/page-layout/page-layout.tsx
```diff
@@ -1,16 +1,16 @@
 import { type FC } from 'react';
 
+import { Header, RouterOutlet } from '~/libs/components/components.js';
```
PavloSukhinin: Let's import this from /header/header.js and router/router.js

PavloSukhinin (changes requested):

frontend/src/libs/components/page-layout/page-layout.tsx
h0wter: Should I add a PageLayout export from components.js? It wasn't exported there before and I figured I could import Header and RouterOutlet from there since there is no circular dependency.

PavloSukhinin (approved):

kalinkaaaa14 (approved):

PR #157, opened 2023-09-03
th-23: Protected routing [BE]

h0wter:
- Added the ability to pass an array to `authStrategy`.
- Now `Controller` accepts default strategies and applies them to all routes of the controller. 
- Strategies can be overridden or canceled for a particular route by passing the required strategies or `null` to `authStrategy`.


h0wter:
Modified verifyJwtStrategy to create a strategy that injects the user if the token exists but does not return an error if the token is missing.

DmytroVoronkov (approved):

Some14u (approved):
Everything seems ok

Druxight (approved):

kalinkaaaa14 (approved):

PR #173, opened 2023-09-04
th-85: Checking on existing business before creating user

TetianaDoroshko:
The issues that resolve this bugfix:
While creating business user there are creating a user at first, recieve its ID fron db, then creating a busines connecting with user. But in this case we've encountered to the problem: if company name or taxNumber is the same to the existed one, BE send request with error but the user is already existed in db.  So he can't change company name and use sign-up form again.
1. That is why there was added the checking on the unique tax number before creating the entities user and busines. Added one more option of response - "Busines with specified tax number already exists".
2. The checking on company name uniqueness was removed and removed corresponding index in db.
3. Also was refactored method findByPhoneOrEmail for reducing a db calls.


TetianaDoroshko:
Wrong branch name(((

PR #182, opened 2023-09-05
th-171: Scroll Bar styling

AlexVlaso:
Снимок экрана 2023-09-04 в 21 08 07


DmytroVoronkov (approved):

albsbz (approved):

frontend/src/assets/css/scroll-bar.scss
```diff
@@ -0,0 +1,18 @@
+/* Works on Firefox */
+html {
+  scrollbar-color: #3563ad rgba(238 239 242 / 50%);
```
h0wter: Maybe we should put these colors in `vars.scss`? Especially `#3563ad` is repeated twice here.

h0wter (changes requested):

frontend/src/assets/css/scroll-bar.scss
AlexVlaso: I thought about it, but we already have 6 types of blue, and the same with gray. In our layout, we use these colors just here. I think we should not clutter vars with non-reusable variables

PR #95, opened 2023-08-26
th-81: Create order card component

mV22110702:

PR #192, opened 2023-09-05
th-191: Fix order card locations displacement

mV22110702:


kalinkaaaa14:
why didn't you fix it in this pr related to order card component, as it's still open ? Now it's hard to check because it's almost repeating all changes from that pr
https://github.com/BinaryStudioAcademy/bsa-2023-towhub/pull/95

PR #194, opened 2023-09-05
th-183: Update validation rule and error message

h0wter:
Fixes phone validation rules to match documentation.


DmytroVoronkov (approved):

kalinkaaaa14 (dismissed):

shared/src/packages/users/libs/validation-schemas/enums/user-validation-rule.enum.ts
```diff
@@ -1,7 +1,7 @@
 const UserValidationRule = {
   PASSWORD: /^(?=.*[A-Za-z])(?=.*\d)[\dA-Za-z]{6,20}$/,
-  NAME: /^[A-Za-z][\s'A-Za-z-]{0,39}$/,
-  PHONE: /^\+\d{8,19}$/,
+  NAME: /^['A-Za-z-]{1,40}$/,
```
albsbz, line 3: This will eliminate check for first symbol to be letter

AlexVlaso (approved):

MaxHopaytsa (approved):

kalinkaaaa14 (approved):

Druxight (approved):

shared/src/packages/users/libs/validation-schemas/enums/user-validation-rule.enum.ts
h0wter: Why do we need this? In that case, `-` and `'` would be considered a valid name.

PavloSukhinin (approved):

PR #197, opened 2023-09-05
th-180: Get current user

TetianaDoroshko:
Created the endpoint for getting user data on BE.
Implement updare user data on the client side.


backend/src/packages/auth/auth.service.ts
```diff
@@ -187,6 +187,20 @@ class AuthService {
     };
   }
 
+  public async getCurrent(
```
Some14u, line 218: As for now this looks more like a "attachUserBusiness" method. Although, I understand why you're doing like this.

I believe, that in the future all getUserBy... methods of userService will be returning user with optional business attached, thus AuthStrategy.INJECT_USER also would have it returning automatically.

frontend/src/libs/components/router/router.tsx
```diff
@@ -1,40 +1,50 @@
...
-      <Route path={AppRoute.SIGN_UP} element={<Auth />} />
-    </Route>
-    <Route
-      path={AppRoute.ROOT}
-      element={<ProtectedRoute allowedUserGroup={UserGroupKey.BUSINESS} />}
-    >
+const Router = (): JSX.Element => {
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
```
Some14u: It looks like you're applying this strategy to any possible session, right?
What if the current application user does not have account yet and wants to use the application anonymously?

I think we should apply this only if there is a jwt key in the localstorage. You should add this check here, and maybe you would consider to wrap it into a local hook (by local i mean it should rest in router/libs/hooks), just to make it look more concise.

Another option might be to apply this only for "protected" routes, however it looks like we don't have frontend protected functionality in public yet. Do you know, who is responsive for that?

frontend/src/packages/auth/auth-api.ts
```diff
@@ -60,6 +60,23 @@ class AuthApi extends HttpApi {
...
+
+  public async getCurrentUser(): Promise<
+    CustomerSignUpResponseDto | BusinessSignUpResponseDto
+  > {
+    const response = await this.load(
+      this.getFullEndpoint(AuthApiPath.CURRENT, {}),
+      {
+        method: 'GET',
+        contentType: ContentType.JSON,
+        hasAuth: true,
```
Some14u, line 72: this hasAuth means, that localstorage is expected to have jwt token from the previous session.
Check the behavior by clearing the localstorage or by opening the app in private tab of the browser.

Some14u (changes requested):

frontend/src/libs/components/router/router.tsx
TetianaDoroshko: Good point, thank you

DmytroVoronkov (approved):

Some14u (approved):

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -50,6 +50,16 @@ const { reducer, actions, name } = createSlice({
...
     });
+    builder.addCase(getCurrent.pending, (state) => {
+      state.dataStatus = DataStatus.PENDING;
+    });
+    builder.addCase(getCurrent.fulfilled, (state, action) => {
+      state.user = action.payload;
+      state.dataStatus = DataStatus.FULFILLED;
+    });
+    builder.addCase(getCurrent.rejected, (state) => {
+      state.dataStatus = DataStatus.REJECTED;
```
Druxight, line 61: We should cleanup storage end logout user in this case

Druxight (changes requested):

frontend/src/libs/hooks/use-get-current-user/use-get-current-user.hook.ts
```diff
@@ -0,0 +1,24 @@
+import { LocalStorage, StorageKey } from '~/libs/packages/storage/storage.js';
+import { actions as authActions } from '~/slices/auth/auth.js';
+
+import { useAppDispatch } from '../hooks.js';
```
PavloSukhinin:
```suggestion
import { useAppDispatch } from '../use-app-dispatch/use-app-dispatch.hook.js';
```
because useGetCurrentUser exported from the same hooks.js file

PavloSukhinin (changes requested):

PavloSukhinin (approved):

Druxight (approved):

PR #210, opened 2023-09-07
th-199: Bug: Missing design elements on the Sign up pages (both Business and client)

TetianaDoroshko:
There was a problem with transforming svg-file during building. As I found out it has to be resolved with replacing absolute path to relative one. In my inctance of progect file has successfully created in build directory.


DmytroVoronkov (approved):

frontend/src/pages/auth/styles.module.scss
```diff
@@ -2,10 +2,10 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  background-color: #3563e9;
   min-height: 100vh;
-  background-image: url('src/assets/img/auth-page-bg.svg');
-  background-size: cover;
+  background-color: #3563e9;
```
h0wter:
```suggestion
  background-color: $blue-dark;
```

h0wter (changes requested):

frontend/src/pages/auth/styles.module.scss
TetianaDoroshko: I believed that it will be changed by props in PageLayout component. But now we've clarified this moment. I've corrected.

Druxight (approved):

PR #229, opened 2023-09-12
th-217: See possible trucks for assigning on a shift

mV22110702:


backend/src/packages/trucks/truck.controller.ts
```diff
@@ -30,17 +30,17 @@ import { type TruckService } from './truck.service.js';
...
- *             - daf
- *             - man
- *             - volvo
- *             - renault
- *             - iveco
- *             - scania
- *             - kenworth
- *             - ford
- *           example: kenworth
+ *             - Miller_industries
```
h0wter: What is the purpose of changing the case? The frontend will have to handle it anyway to remove the _, and we can change the case of the first letter via css.

frontend/src/libs/packages/socket/socket.package.ts
```diff
@@ -1,38 +1,64 @@
...
   private io: Socket | undefined;
 
-  public connect(): void {
+  public connect(user: UserEntityObjectWithGroupT | null): void {
+    const auth = user ? { userId: user.id } : {};
     this.io = io(config.ENV.API.SERVER_URL, {
       transports: ['websocket', 'polling'],
+      auth,
     });
   }
```
albsbz, line 36:
```
type ServerToClientEvents = {
  [ServerSocketEvent.name]: () => void;
  basicEmit: (a: number, b: string, c: Buffer) => void;
  withAck: (d: string, callback: (error: number) => void) => void;
};

type ClientToServerEvents = {
  hello: () => void;
};

class SocketService {
  private io: Socket<ServerToClientEvents, ClientToServerEvents> | undefined;

  public connect(): void {
    const socket: Socket<ServerToClientEvents, ClientToServerEvents> = io(
      config.ENV.API.SERVER_URL,
      {
        transports: ['websocket', 'polling'],
      },
    );
    this.io = socket;
  }
 public addListener(
    event: keyof ServerToClientEvents,
    listener: () => void,
  ): void {
    this.io?.on(event, listener);
  }

```
Maybe we can pass ServerToClientEvents and ClientToServerEvents as generics like this (like in socket.io docs)

mV22110702: 1. Wouldn't we lose listener parameters by specifying its type as () => void?
2. If we have to specify a value in eventPayload, then the fields from Union type of all payloads will be offered

I suggest we leave the refactoring for now, as it will take a lot of time, and just leave the Generic we have for now, especially since the typing when specifying a specific event name works.

mV22110702: Also let's make it in another ticket to merge the main functionality sooner.

backend/src/libs/packages/socket/socket.service.ts
```diff
@@ -42,6 +104,75 @@ class SocketService {
...
+        }
+        await socketEndShift({
+          io: this.io,
+          startedShiftsStore,
+          truckService,
+          shiftService,
+          user,
+        });
+      });
     });
```
albsbz, line 95: It seems to me it's better to inject socket service in truckService,  shiftService, userService, where we need and keep socket service as thin as possible

backend/src/packages/shifts/shift.socket-service.ts
```diff
@@ -0,0 +1,164 @@
...
+        this.socket.emit(ClientSocketEvent.DRIVER_TIMED_OUT);
+        void socketEndShift({
+          io: this.io,
+          startedShiftsStore: this.startedShiftsStore,
+          shiftService: this.shiftService,
+          truckService: this.truckService,
+          user,
+        });
+      },
+      HOUR_IN_MS,
```
PavloSukhinin: not sure why we want to end shift after 1 hour 🤔

backend/src/packages/trucks/truck.service.ts
```diff
@@ -18,6 +32,32 @@ class TruckService implements IService {
...
+
+    if (!usersTruckRecords) {
+      return [];
+    }
+
+    return await Promise.all(
+      usersTruckRecords.map(async (userTruckRecord) => {
+        const [truck = null] = await this.repository.findById(
+          userTruckRecord.truckId,
+        );
```
PavloSukhinin: Let's do it in 1 db call. Promise.all can easily exhaust database connection pool.

backend/src/packages/users-trucks/users-trucks.service.ts
```diff
@@ -0,0 +1,79 @@
...
+      userId,
+    });
+
+    if (usersTrucksRecords.length === 0) {
+      return [];
+    }
+
+    return usersTrucksRecords.map((usersTrucksRecord) =>
+      UsersTrucksEntity.initialize(usersTrucksRecord).toObject(),
+    );
```
PavloSukhinin:
```suggestion
    return usersTrucksRecords.map((usersTrucksRecord) => {
      return UsersTrucksEntity.initialize(usersTrucksRecord).toObject()
     },
    );
```

frontend/src/libs/packages/store/middlewares/socket/socket.middleware.ts
```diff
@@ -0,0 +1,57 @@
...
+  ReturnType<ReturnType<typeof useAppDispatch>>,
+  RootState,
+  ReturnType<typeof useAppDispatch>
+> = ({ dispatch }) => {
+  return (next) => (action: ReturnType<ActionCreatorWithPayload<unknown>>) => {
+    if (action.type === driverActions.endShift.type) {
+      socketClient.emit({ event: ServerSocketEvent.END_SHIFT });
+    } else if (action.type === driverActions.startShift.type) {
+      const { truckId } =
+        action.payload as ServerSocketEventParameter[typeof ServerSocketEvent.START_SHIFT];
```
PavloSukhinin: I'm not sure that it's a good idea to keep all side effects in a single place - it will become cluttered. Can we move socketClient events into actions?

frontend/src/libs/packages/store/middlewares/socket/socket.middleware.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      socketClient.emitWithAck({
+        event: ServerSocketEvent.START_SHIFT,
```
PavloSukhinin: I believe it's promised-based version of emit and it can be awaited `const foo = await socketClient.emitWithAck`

frontend/src/packages/trucks/trucks-api.ts
```diff
@@ -30,7 +34,27 @@ class TruckApi extends HttpApi {
...
+    'userId'
+  >): Promise<GetAllTrucksByUserIdResponseDto> {
+    const response = await this.load(
+      this.getFullEndpoint(TruckApiPath.$USER_ID, {
+        userId: userId.toString(),
+      }),
+      {
+        method: 'GET',
+        contentType: ContentType.JSON,
+        hasAuth: false,
```
PavloSukhinin:
```suggestion
  public async getAllTrucksByUserId({
    userId,
  }: Pick<
    UsersTrucksEntityT,
    'userId'
  >): Promise<GetAllTrucksByUserIdResponseDto> {
    const response = await this.load(
      this.getFullEndpoint(TruckApiPath.$USER_ID, {
        userId: userId.toString(),
      }),
      {
        method: 'GET',
        contentType: ContentType.JSON,
        hasAuth: true,
```
there should be auth 🤔

frontend/src/slices/driver/driver.slice.ts
```diff
@@ -0,0 +1,52 @@
...
+  activeTruckId: number | null;
+  shiftStatus: ShiftStatusValue;
+};
+
+const initialState: State = {
+  truckChoiceStatus: {
+    isSuccess: false,
+    isPending: false,
+    isIdle: false,
+  },
```
PavloSukhinin: Why not just single enum value? Like with DataStatus

frontend/src/slices/trucks/selectors.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const selectTrucks = (state: RootState): GetAllTrucksByUserIdResponseDto =>
+  state.trucks.trucks;
+
+const selectTruck =
+  (truckId: number | null) =>
+  (state: RootState): TruckEntityT | undefined =>
+    truckId
+      ? state.trucks.trucks.items.find((truck) => truck.id === truckId)
+      : undefined;
```
PavloSukhinin: curly + return, please

frontend/src/slices/trucks/selectors.ts
```diff
@@ -0,0 +1,23 @@
...
+const selectDataStatus = (state: RootState): ValueOf<typeof DataStatus> => {
+  return state.trucks.dataStatus;
+};
+
+const selectTrucks = (state: RootState): GetAllTrucksByUserIdResponseDto =>
+  state.trucks.trucks;
+
+const selectTruck =
+  (truckId: number | null) =>
+  (state: RootState): TruckEntityT | undefined =>
```
PavloSukhinin: can it be TruckEntityT | null? I'd prefer to use null instead of undefined because they have different concepts.
Null - stands for No data, while undefined is about 'something is not defined yet'

shared/src/libs/packages/socket/libs/types/client-socket-event-parameter.type.ts
```diff
@@ -0,0 +1,19 @@
+import { type ClientSocketEvent } from '../enums/enums.js';
+
+type ClientSocketEventParameter = {
+  [ClientSocketEvent.TRUCK_CHOSEN]: {
+    truckId: number;
+  };
+  [ClientSocketEvent.TRUCK_AVAILABLE]: {
+    truckId: number;
+  };
```
PavloSukhinin: create a type for `truckId: number;`

shared/src/packages/trucks/libs/enums/truck-status.enum.ts
```diff
@@ -0,0 +1,8 @@
+const TruckStatus = {
+  NOT_AVAILABLE: 'Not available',
```
PavloSukhinin: it's only for UI, or will be used on backend?

PavloSukhinin (changes requested):

shared/src/packages/trucks/libs/enums/truck-status.enum.ts
mV22110702: Particular `Not available` is not used on backend, so it will be deleted soon

kalinkaaaa14 (approved):

Druxight (approved):

PR #264, opened 2023-09-15
th-243: Add possibility accept/decline order

MaxHopaytsa:
- Endpoint added 
- Added methods to the controller/service
- Added api/action/slice on the client


backend/src/packages/orders/order.service.ts
```diff
@@ -216,7 +225,108 @@ class OrderService implements Omit<IService, 'find'> {
...
+    };
+
+    const order = OrderEntity.initialize(orderExtended).toObject();
+
+    this.socketService.notifyOrderUpdate(order.id, order);
+
+    return order;
+  }
+
+  public async updateAcceptStatus(parameters: {
```
PavloSukhinin:
```suggestion
  public async updateAcceptStatus({ id, payload, user }: {
```

backend/src/packages/orders/order.service.ts
```diff
@@ -216,7 +225,108 @@ class OrderService implements Omit<IService, 'find'> {
...
+    id: OrderEntityT['id'];
+    payload: OrderUpdateAcceptStatusRequestDto;
+    user: UserEntityObjectWithGroupT | null;
+  }): Promise<OrderResponseDto> {
+    let acceptStatus: OrderStatusValues = OrderStatus.PENDING;
+
+    if (parameters.user && parameters.user.group.key === UserGroupKey.DRIVER) {
```
PavloSukhinin: let's introduce a helper 
```
const checkIsDriver = (groupKey: UserGroupKeyT): groupKey is typeof UserGroupKey.DRIVER
```

PavloSukhinin (changes requested):

backend/src/libs/packages/database/generated-schema/0012_add_refused_in_order_statur.sql
Druxight, line 1: typo in the filename, statur instead of status

backend/src/packages/orders/order.service.ts
```diff
@@ -216,7 +227,106 @@ class OrderService implements Omit<IService, 'find'> {
...
 
-    return OrderEntity.initialize(orderExtended).toObject();
+    const order = OrderEntity.initialize(orderExtended).toObject();
+
+    this.socketService.notifyOrderUpdate(order.id, order);
+
+    return order;
+  }
+
+  public async updateByCustomer({
```
Druxight: We will set the driver to the order from the Driver side, not the Customer side. 
Code is correct but I suppose we should rename the function name
Correct me if I wrong

backend/src/packages/orders/order.service.ts
```diff
@@ -216,7 +227,106 @@ class OrderService implements Omit<IService, 'find'> {
...
+      await this.shiftService.checkDriverStartShift(user.id);
+
+      return await this.update({
+        id,
+        payload: { status: acceptStatus },
+        user,
+      });
+    }
+
+    if (checkIsCustomer(user?.group.key as UserGroupKeyT)) {
```
Druxight: Only Drivers can accept/decline orders. The customer has access to the create op.

frontend/src/libs/packages/middleware/socket.middleware.ts
```diff
@@ -0,0 +1,56 @@
...
+        void dispatch(updateOrderFromSocket(order));
+      },
+    );
+  }
+
+  return (next: Dispatch) => (action: AnyAction) => {
+    if (listenOrderUpdates.type === action.type && socketInstance) {
+      socketInstance.emit<typeof ClientSocketEvent.SUBSCRIBE_ORDER_UPDATES>(
+        ClientSocketEvent.SUBSCRIBE_ORDER_UPDATES,
+        { orderId: `${action.payload as string}` },
```
Druxight:
```suggestion
        { orderId: action.payload as string },
```

frontend/src/packages/orders/orders-api.ts
```diff
@@ -53,6 +54,23 @@ class OrdersApi extends HttpApi {
...
+    orderId: string,
+    payload: OrderUpdateAcceptStatusRequestDto,
+  ): Promise<OrderResponseDto> {
+    const response = await this.load(
+      this.getFullEndpoint(OrdersApiPath.$ID, { id: orderId }),
+      {
+        method: 'PATCH',
+        contentType: ContentType.JSON,
+        payload: JSON.stringify(payload),
+        hasAuth: true,
```
Druxight, line 82: Why do we need to pass the 'hasAuth' to the server?
Such an approach is wrong.
If we need to have an API endpoint that has Auth conditionally, we should create 2 separate endpoints. 1 with auth and 1 without

frontend/src/slices/orders/libs/enums/action-names.enum.ts
```diff
@@ -0,0 +1,12 @@
+import { name as sliceName } from '~/slices/orders/order.slice.js';
+
+const ActionNames = {
+  CHANGE_ACCEPT_ORDER_STATUS: `${sliceName}/vhange-accept-order-status`,
+  SOCKET: {
+    UPDATE_ORDER: `${sliceName}/socket/update-order`,
+    LISTEN_ORDER_UPDATES: `${sliceName}/listen-order-updates`,
+    STOP_LISTEN_ORDER_UPDATES: `${sliceName}/stop-listen-order-updates`,
```
Druxight: Please unify the naming here. Remove the 'socket' part from the UPDATE_ORDER or add it everywhere

frontend/src/slices/orders/order.slice.ts
```diff
@@ -1,13 +1,18 @@
...
+import {
+  calculateOrderPrice,
+  changeAcceptOrderStatus,
+  createOrder,
+  updateOrderFromSocket,
+} from './actions.js';
 
 type State = {
-  orders: OrderResponseDto[];
+  orders: (OrderResponseDto | undefined)[];
```
Druxight: I do not think that we want to work with an array of undefined.

shared/src/libs/packages/socket/libs/enums/server-socket-event.enum.ts
```diff
@@ -1,7 +1,7 @@
 const ServerSocketEvent = {
   CONNECTION: 'connection',
   DISCONNECT: 'disconnect',
-  TRUCK_LOCATION_UPDATE: 'truck_location_update',
```
Druxight: I think it will be used in other PRs, please rollback

shared/src/packages/orders/libs/enums/order-status.enum.ts
```diff
@@ -2,6 +2,7 @@ const OrderStatus = {
   PENDING: 'pending',
   CONFIRMED: 'confirmed',
   CANCELED: 'canceled',
+  REFUSED: 'refused',
```
Druxight: REJECTED

Druxight (changes requested):

frontend/src/packages/orders/orders-api.ts
PavloSukhinin: It's an instruction for a frontend Http service to use bearer auth.

frontend/src/slices/orders/libs/enums/action-names.enum.ts
```diff
@@ -0,0 +1,12 @@
+import { name as sliceName } from '~/slices/orders/order.slice.js';
+
+const ActionNames = {
```
PavloSukhinin: ActionName

backend/src/packages/orders/order.service.ts
```diff
@@ -301,6 +411,21 @@ class OrderService implements Omit<IService, 'find'> {
...
+  ): OrderStatusValues {
+    if (user && checkIsDriver(user.group.key)) {
+      return isAccepted ? OrderStatus.CONFIRMED : OrderStatus.REFUSED;
+    } else if ((!user || checkIsCustomer(user.group.key)) && !isAccepted) {
+      return OrderStatus.CANCELED;
+    } else {
+      throw new NotFoundError({
+        message: HttpMessage.USER_CAN_NOT_ACCEPT_OR_DECLINE_ORDER,
+      });
+    }
```
PavloSukhinin: I believe it can be interpreted like this
```
if (checkIsDriver(user?.group.key ?? null)) {
  return...
}

if (...) {
  return OrderStatus.CANCELED;
}

throw new NotFoundError({})
```

frontend/src/slices/orders/libs/enums/action-names.enum.ts
kalinkaaaa14: also don't forget to rename file

shared/src/libs/packages/socket/libs/enums/room-prefixes.enum.ts
```diff
@@ -0,0 +1,5 @@
+const RoomPrefixes = {
```
kalinkaaaa14: we already have RoomPrefix enum in shared, reuse

frontend/src/packages/orders/orders-api.ts
```diff
@@ -60,12 +62,47 @@ class OrdersApi extends HttpApi {
...
+    payload: OrderUpdateAcceptStatusRequestDto,
+  ): Promise<OrderUpdateAcceptStatusResponseDto> {
+    const response = await this.load(
+      this.getFullEndpoint(OrdersApiPath.DRIVER, { orderId }),
+      {
+        method: 'PATCH',
+        contentType: ContentType.JSON,
+        payload: JSON.stringify(payload),
+
+        hasAuth: true,
```
PavloSukhinin:
```suggestion
        payload: JSON.stringify(payload),
        hasAuth: true,
```

backend/src/libs/packages/socket/socket.service.ts
```diff
@@ -155,6 +156,15 @@ class SocketService {
       .emit(ServerToClientEvent.ORDER_UPDATED, order);
   }
 
+  public notifyOrderUpdateStatus(
```
Druxight: why do we need a new event if we already have the ServerToClientEvent.ORDER_UPDATED

PavloSukhinin (approved):

Druxight (approved):

PR #269, opened 2023-09-15
th-236: Business Journey: Saving Order with shift id + getting all orders with joined driver & truck info

TetianaDoroshko:
1. Changed order schema: add shiftId column, change relation to driverId
2. Orders responses now contain joined info about driver id and truck id (saved in the shift record)
3. Replace driverId to truckId that is supposed to be in create order request
4. While creating order there is checking for open shift for specified truck.


backend/src/packages/orders/order.entity.ts
```diff
@@ -80,13 +102,16 @@ class OrderEntity implements IEntity {
...
       customerName,
       customerPhone,
+      shift,
+      driver,
+      truck,
     });
   }
 
-  public static initializeNew({
+  public static initializeDb({
```
PavloSukhinin: Let's keep `initializeNew` name to keep consistency with other entities

backend/src/packages/orders/order.entity.ts
```diff
@@ -80,13 +102,16 @@ class OrderEntity implements IEntity {
...
+    id,
```
PavloSukhinin: Not sure why we pass `id` into initializeNew

backend/src/packages/orders/order.entity.ts
```diff
@@ -22,12 +29,21 @@ class OrderEntity implements IEntity {
...
   private customerName: OrderEntityT['customerName'] | null;
 
   private customerPhone: OrderEntityT['customerPhone'] | null;
 
+  private shift: {
+    id: ShiftEntity['id'];
+  };
+
+  private driver:
+    | (Pick<UserEntityT, 'id' | 'firstName' | 'lastName' | 'email' | 'phone'> &
```
kalinkaaaa14: DriverCommonDetails & UserCommonDetails - can we reuse them?

shared/src/packages/orders/libs/types/order-response-dto.type.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type OrderResponseDto = Omit<OrderEntity, 'shift' | 'driver' | 'truck'> & {
+  shift: {
+    id: ShiftEntity['id'];
+    driver:
+      | (Pick<
+          UserEntityT,
+          'id' | 'firstName' | 'lastName' | 'email' | 'phone'
+        > &
+          Pick<DriverEntity, 'driverLicenseNumber'>)
```
kalinkaaaa14: same as above, we need to make readable type for this
UserCommonDetails & DriverCommonDetails

shared/src/packages/orders/libs/types/order-entity.type.ts
```diff
@@ -9,10 +13,15 @@ type OrderEntity = {
...
   userId: number | null;
-  businessId: number;
-  driverId: number;
+  businessId: number | null;
   customerName: string | null;
   customerPhone: string | null;
+  shift: { id: number };
+  driver:
+    | (Pick<UserEntityT, 'id' | 'firstName' | 'lastName' | 'email' | 'phone'> &
+        Pick<DriverEntity, 'driverLicenseNumber'>)
```
kalinkaaaa14: same about readable type

shared/src/packages/orders/libs/types/order-entity.type.ts
```diff
@@ -9,10 +13,15 @@ type OrderEntity = {
   endPoint: string;
   status: OrderStatusValues;
   userId: number | null;
-  businessId: number;
-  driverId: number;
+  businessId: number | null;
   customerName: string | null;
   customerPhone: string | null;
+  shift: { id: number };
```
kalinkaaaa14: why it's object? 
either shiftId: number or shift: ShiftEntity

shared/src/packages/orders/libs/types/order-create-response-dto.type.ts
```diff
@@ -1,5 +1,11 @@
 import { type OrderEntity } from './order-entity.type.js';
 
-type OrderCreateResponseDto = Omit<OrderEntity, 'userId'>;
+type OrderCreateResponseDto = Omit<OrderEntity, 'userId'> & {
+  // shift: {
```
kalinkaaaa14: remove comments

backend/src/packages/orders/order.repository.ts
```diff
@@ -24,72 +27,111 @@ class OrderRepository implements Omit<IRepository, 'find'> {
...
+        scheduledTime: this.ordersSchema.scheduledTime,
+        startPoint: this.ordersSchema.startPoint,
+        endPoint: this.ordersSchema.endPoint,
+        status: this.ordersSchema.status,
+        carsQty: this.ordersSchema.carsQty,
+        customerName: this.ordersSchema.customerName,
+        customerPhone: this.ordersSchema.customerPhone,
+        shift: {
+          id: this.ordersSchema.shiftId,
+        },
```
PavloSukhinin:
```suggestion
private trucksSchema: DatabaseSchema['trucks'];

  private usersSchema: DatabaseSchema['users'];

  private shiftsSchema: DatabaseSchema['shifts'];

  private driversSchema: DatabaseSchema['drivers'];

  public constructor(
    database: Pick<IDatabase, 'driver'>,
    {
      orders,
      trucks,
      users,
      shifts,
      drivers,
    }: Pick<
      DatabaseSchema,
      'orders' | 'users' | 'trucks' | 'shifts' | 'drivers'
    >,
  ) {
    this.db = database;
    this.ordersSchema = orders;
    this.trucksSchema = trucks;
    this.usersSchema = users;
    this.shiftsSchema = shifts;
    this.driversSchema = drivers;
  }

  public async findById(id: OrderEntityT['id']): Promise<OrderEntityT | null> {
    const [order = null] = await this.db
      .driver()
      .select({
        id: this.ordersSchema.id,
        userId: this.ordersSchema.userId,
        businessId: this.ordersSchema.businessId,
        price: this.ordersSchema.price,
        scheduledTime: this.ordersSchema.scheduledTime,
        startPoint: this.ordersSchema.startPoint,
        endPoint: this.ordersSchema.endPoint,
        status: this.ordersSchema.status,
        carsQty: this.ordersSchema.carsQty,
        customerName: this.ordersSchema.customerName,
        customerPhone: this.ordersSchema.customerPhone,
        shift: {
          id: this.ordersSchema.shiftId,
        },
        driver: {
          id: this.shiftsSchema.driverId,
          firstName: this.usersSchema.firstName,
          lastName: this.usersSchema.lastName,
          email: this.usersSchema.email,
          phone: this.usersSchema.phone,
          driverLicenseNumber: this.driversSchema.driverLicenseNumber,
        },
        truck: {
          id: this.shiftsSchema.truckId,
          licensePlateNumber: this.trucksSchema.licensePlateNumber,
        },
      })
      .from(this.ordersSchema)
      .innerJoin(
        this.shiftsSchema,
        eq(this.ordersSchema.shiftId, this.shiftsSchema.id),
      )
      .innerJoin(
        this.usersSchema,
        eq(this.shiftsSchema.driverId, this.usersSchema.id),
      )
      .innerJoin(
        this.driversSchema,
        eq(this.driversSchema.userId, this.shiftsSchema.driverId),
      )
      .innerJoin(
        this.trucksSchema,
        eq(this.shiftsSchema.truckId, this.trucksSchema.id),
      )
      .where(eq(this.ordersSchema.id, id));

    return order;
  }
```

backend/src/packages/orders/order.repository.ts
```diff
@@ -24,72 +27,111 @@ class OrderRepository implements Omit<IRepository, 'find'> {
...
+          id: schema.shifts.driverId,
+          firstName: schema.users.firstName,
+          lastName: schema.users.lastName,
+          email: schema.users.email,
+          phone: schema.users.phone,
+          driverLicenseNumber: schema.drivers.driverLicenseNumber,
+        },
+        truck: {
+          id: schema.shifts.truckId,
+          licensePlateNumber: schema.drivers.driverLicenseNumber,
```
PavloSukhinin: should be licensePlateNumber, not driverLicenseNumber

shared/src/packages/orders/libs/types/order-create-request-dto.type.ts
```diff
@@ -2,7 +2,14 @@ import { type OrderEntity } from './order-entity.type.js';
 
 type OrderCreateRequestDto = Omit<
```
kalinkaaaa14: maybe using pick will be shorter?

backend/src/packages/shifts/shift.repository.ts
```diff
@@ -40,6 +43,55 @@ class ShiftRepository implements IRepository {
...
+    | (ShiftDatabaseModel & { businessId: number; driverLicenseNumber: string })
+    | null
+  > {
+    const [shift = null] = await this.db
+      .driver()
+      .select()
+      .from(this.shiftSchema)
+      .innerJoin(
+        schema.drivers,
+        eq(this.shiftSchema.driverId, schema.drivers.userId),
```
PavloSukhinin, line 76:
```suggestion
        eq(this.shiftSchema.driverId, this.driversSchema.userId),
```
and change it in all other places

backend/src/packages/shifts/shift.repository.ts
```diff
@@ -40,6 +43,55 @@ class ShiftRepository implements IRepository {
...
+          isNull(this.shiftSchema.endDate),
+          eq(this.shiftSchema.truckId, truckId),
+        ),
+      )
+      .execute();
+
+    return shift
+      ? {
+          ...shift.shifts,
+          businessId: shift.driver_details.businessId,
```
PavloSukhinin, line 89: not sure why it's in snake_case

shared/src/libs/packages/http/libs/enums/http-error-message.enum.ts
```diff
@@ -22,6 +22,8 @@ const HttpMessage = {
   SHIFT_ALREADY_CLOSED: 'Specified shift is already closed',
   NOT_ACCESS: 'You do not have enough rights to operate this shift',
   ORDER_DOES_NOT_EXIST: 'Order does not exist!',
+  SHIFT_NOT_OPEN: 'This truck is not working now',
+  TRUCK_DOES_NOT_EXISTS: 'Truck does not already exist',
```
PavloSukhinin:
```suggestion
  TRUCK_DOES_NOT_EXISTS: 'Truck does not exist',
```
🤔

shared/src/libs/packages/http/libs/enums/http-error-message.enum.ts
```diff
@@ -22,6 +22,8 @@ const HttpMessage = {
   SHIFT_ALREADY_CLOSED: 'Specified shift is already closed',
   NOT_ACCESS: 'You do not have enough rights to operate this shift',
   ORDER_DOES_NOT_EXIST: 'Order does not exist!',
+  SHIFT_NOT_OPEN: 'This truck is not working now',
```
PavloSukhinin:
```suggestion
  SHIFT_NOT_OPEN: 'This truck is inactive, please choose another one',
```

shared/src/packages/orders/libs/types/order-create-response-dto.type.ts
```diff
@@ -1,5 +1,11 @@
...
+  //   id: OrderEntity['shiftId'];
+  //   driverId: OrderEntity['driverId'];
+  //   truckId: OrderEntity['truckId'];
+  // };
+};
```
PavloSukhinin: remove commented code, please

shared/src/packages/orders/libs/types/order-response-dto.type.ts
PavloSukhinin: Let's introduce a name for this type. And use it here, and here `shared/src/packages/orders/libs/types/order-entity.type.ts`.
And also here `backend/src/packages/orders/order.entity.ts`

PavloSukhinin (changes requested):
Please update schema in readme.md

shared/src/packages/orders/libs/types/order-entity.type.ts
TetianaDoroshko: Because later I am pitting into this object driver and truck for response. But ok, I'll change it.

backend/src/packages/orders/order.repository.ts
Some14u:
```
      .innerJoin(
        this.driversSchema,
        eq(this.driversSchema.userId, this.shiftsSchema.driverId),
      )
```
      
      better be written as
      
```
      .innerJoin(
        this.driversSchema,
        eq(this.shiftsSchema.driverId, this.driversSchema.userId),
      )
```
      just for consistency. I was stuck looking at this in the beginning :)

shared/src/packages/orders/libs/types/order-create-request-dto.type.ts
TetianaDoroshko: it will be the same length

backend/src/packages/orders/order.entity.ts
```diff
@@ -22,12 +29,21 @@ class OrderEntity implements IEntity {
...
 
-  private driverId: OrderEntityT['driverId'];
-
   private customerName: OrderEntityT['customerName'] | null;
 
   private customerPhone: OrderEntityT['customerPhone'] | null;
 
+  private shift: {
+    id: ShiftEntity['id'];
+  };
```
mV22110702: Maybe it will be better to use Pick<ShiftEntity,'id'> instead?

```suggestion
  private shift: Pick<ShiftEntity,'id'>;
```

backend/src/packages/orders/order.entity.ts
```diff
@@ -95,12 +120,14 @@ class OrderEntity implements IEntity {
...
+    shiftId,
     customerName,
     customerPhone,
-  }: Omit<OrderEntityT, 'id'>): OrderEntity {
+  }: Omit<OrderEntityT, 'shift' | 'driver' | 'truck'> & {
+    shiftId: number;
+  }): OrderEntity {
     return new OrderEntity({
-      id: null,
+      id,
```
mV22110702: To Pavlo's comment on necessity of passing id.
```suggestion
      id: null
```

backend/src/packages/orders/order.service.ts
```diff
@@ -24,168 +25,232 @@ class OrderService implements Omit<IService, 'find'> {
...
+      throw new NotFoundError({ message: HttpMessage.TRUCK_DOES_NOT_EXISTS });
+    }
+    const shift =
+      await this.shiftService.findOpenedByTruckWithBusiness(truckId);
 
-    if (!foundOrder) {
+    if (!shift) {
       throw new NotFoundError({
-        message: HttpMessage.ORDER_DOES_NOT_EXIST,
+        message: HttpMessage.SHIFT_NOT_OPEN,
```
mV22110702, line 88:
```suggestion
        message: HttpMessage.SHIFT_NOT_OPENED,
```

backend/src/packages/orders/order.entity.ts
TetianaDoroshko: it is not actually new, the data comes from db and have an id. At first, i named these methods initializeDb() and toObjectDb().
However, IEntity expects to has toNewObject(), but there aren't needing in this method at this entity. So I named these methods initializeNew(), toNewObject().

backend/src/packages/shifts/shift.repository.ts
TetianaDoroshko: Because if I don't define the select() statement, drizzle returns the result of join request like an object with nested objects named as a corresponding table in db.

backend/src/packages/orders/order.repository.ts
```diff
@@ -24,72 +27,111 @@ class OrderRepository implements Omit<IRepository, 'find'> {
...
   }
 
-  public async create(entity: OrderEntity): Promise<OrderEntity> {
-    const [result] = await this.db
+  public async create(
+    entity: Omit<OrderEntityT, 'id' | 'shift' | 'driver' | 'truck'> & {
+      shiftId: number;
+    },
+  ): Promise<OrderDatabaseModel | null> {
+    const [result = null] = await this.db
```
mV22110702: It can be either error, if some DB constraint is violated, or created DB record. In case of order schema, there're no constraints, so there can't be such an error. Hence, it is guaranteed to return inserted record. Also, other repositories' create method do not have an option of returning null as a result.
```suggestion
  ): Promise<OrderDatabaseModel> {
    const [result] = await this.db
```

backend/src/packages/orders/order.service.ts
```diff
@@ -24,168 +25,232 @@ class OrderService implements Omit<IService, 'find'> {
...
+      endPoint,
+      status: OrderStatus.PENDING,
+      userId,
+      businessId: shift.businessId,
+      shiftId: shift.id,
+      customerName,
+      customerPhone,
+    });
+
+    if (!order) {
```
mV22110702: If we remove null from orderRepository.create return types, there will be no need in this if.

backend/src/packages/orders/order.service.ts
```diff
@@ -24,168 +25,232 @@ class OrderService implements Omit<IService, 'find'> {
...
     }
 
-    const { order, driver } = result;
-
-    const foundOrder = { ...order.toObject(), driver };
-
     if (user?.group.key === UserGroupKey.BUSINESS) {
       const business = await this.businessService.findByOwnerId(user.id);
 
       if (business) {
```
mV22110702:
```suggestion
      if (!business) {
       throw new NotFoundError({
        message: HttpMessage.BUSINESS_DOES_NOT_EXIST,
      });
        }
```

mV22110702: If we immediately throw if user's business does not exist, that streamlines flow and makes code more understandable.

backend/src/packages/shifts/shift.repository.ts
```diff
@@ -40,6 +43,55 @@ class ShiftRepository implements IRepository {
...
+      )
+      .where(
+        and(
+          isNull(this.shiftSchema.endDate),
+          eq(this.shiftSchema.truckId, truckId),
+        ),
+      )
+      .execute();
+
+    return shift
```
mV22110702, line 86: May be it's better to add if(!shift){return null;}, and then return transformed object without nesting it into ternary expression?

backend/src/packages/shifts/shift.service.ts
```diff
@@ -66,6 +66,35 @@ class ShiftService implements IService {
...
+
+  public async findOpenedByTruckWithBusiness(
+    truckId: number,
+  ): Promise<
+    (ShiftEntityT & { businessId: number; driverLicenseNumber: string }) | null
+  > {
+    const shiftDatabase =
+      await this.shiftRepository.getOpenedByTruckWithBusiness(truckId);
+
+    if (shiftDatabase) {
```
mV22110702: It'll look simpler if we immediately return null: if(!shiftDatabase){return null;}

shared/src/packages/orders/libs/types/order-create-request-dto.type.ts
```diff
@@ -2,7 +2,14 @@ import { type OrderEntity } from './order-entity.type.js';
...
-  'id' | 'userId' | 'businessId' | 'price' | 'status' | 'driver'
->;
+  | 'id'
+  | 'userId'
+  | 'businessId'
+  | 'price'
+  | 'status'
+  | 'driver'
+  | 'shift'
+  | 'truck'
```
mV22110702: If we keep Omit:
image

So, it's better to make it Pick<>, because:
1. There're only 6 properties to pick instead of 8;
2. It will be clearer on what properties are here, because now, to understand it, we have to omit many properties, and still there'll be a lot of properties left.

shared/src/packages/orders/libs/types/order-response-dto.type.ts
```diff
@@ -0,0 +1,22 @@
+import { type DriverEntity } from '~/packages/drivers/drivers.js';
+import { type ShiftEntity } from '~/packages/shifts/shifts.js';
+import { type TruckEntity } from '~/packages/trucks/trucks.js';
+import { type UserEntityT } from '~/packages/users/users.js';
+
+import { type OrderEntity } from './order-entity.type.js';
+
+type OrderResponseDto = Omit<OrderEntity, 'shift' | 'driver' | 'truck'> & {
+  shift: {
+    id: ShiftEntity['id'];
```
mV22110702, line 10: 1. There's no need to omit 'shift', because if you combine types via '&', then their properties are merged.
2. If we don't omit 'shift', then we should remove  "id: ShiftEntity['id'];"

mV22110702 (changes requested):

backend/src/packages/shifts/shift.repository.ts
TetianaDoroshko: I think 1 return+ternary is better than if+2 returns

backend/src/packages/orders/order.entity.ts
TetianaDoroshko: I've commented this case above

backend/src/packages/shifts/shift.repository.ts
mV22110702: Also true

backend/src/packages/orders/order.entity.ts
```diff
@@ -22,12 +28,16 @@ class OrderEntity implements IEntity {
...
-  private driverId: OrderEntityT['driverId'];
-
   private customerName: OrderEntityT['customerName'] | null;
 
   private customerPhone: OrderEntityT['customerPhone'] | null;
 
+  private shiftId: ShiftEntity['id'];
+
+  private driver: DriverInfo | null;
+
```
albsbz, line 38: why we add driver to the entity if it used only for responses? Entity should correspond to database entity as I see it, the same for the truck field.  If we create new order there is no driver and truck field in it. We can add driver to resposeDto and use it for typing

PavloSukhinin (approved):

Druxight (approved):

PR #272, opened 2023-09-15
th-270: Fix authorized business-user can get data, change data, delete business of any business by id

mV22110702:


MaxHopaytsa (approved):

AlexVlaso (approved):

kalinkaaaa14 (dismissed):

kalinkaaaa14:
actually during creation/updating/deletion - I guess we don't need to pass /:businessId in params and then check if logged in user has access to this operation

for example for creating a new driver - we can refactor it on /business/drivers, and found the businessId by authorized user, so we don't need to check if  logged in user has access to this operation, and don't need to pass param
but we can do it later, now it also will work correctly as you've added checks everywhere

kalinkaaaa14 (dismissed):

shared/src/libs/enums/app-error-message.enum.ts
```diff
@@ -1,6 +1,7 @@
 const AppErrorMessage = {
   INVALID_QUERY: 'Invalid SQL query was provided!',
   ENTITY_NOT_FOUND: 'Requested entity not found!',
+  ENTITY_ACCESS_DENIED: 'The user does not have access to the entity',
   INVALID_FILE_INPUT_CONFIG:
     'Invalid file input config was provided to file validator!',
```
PavloSukhinin:
```suggestion
  INVALID_QUERY: 'Invalid SQL query was provided!',
  ENTITY_NOT_FOUND: 'Requested entity not found!',
  ENTITY_ACCESS_DENIED: 'The user does not have access to the entity!',
  INVALID_FILE_INPUT_CONFIG:
    'Invalid file input config was provided to file validator!',
```
to keep messages in the same format

PavloSukhinin (changes requested):

kalinkaaaa14 (dismissed):

backend/src/packages/business/business.service.ts
```diff
@@ -12,15 +15,19 @@ import {
   type DriverGetAllResponseDto,
 } from '../drivers/drivers.js';
 import { type ShiftEntity } from '../shifts/shift.js';
-import { type TruckEntity } from '../trucks/libs/types/types.js';
+import { type TruckEntityT } from '../trucks/libs/types/types.js';
```
denred: Why did you change "TruckEntity" to "TruckEntityT"?

backend/src/packages/business/business.service.ts
```diff
@@ -42,10 +49,26 @@ class BusinessService implements IService {
...
+      return null;
+    }
+
+    const isOwner = this.checkIsOwner({
+      userId: owner.id,
+      business: foundBusiness,
+    });
+
+    if (!isOwner) {
+      throw new EntityAccessDeniedError({});
```
denred, line 68: Why didn't you specify an error code?

backend/src/packages/business/business.service.ts
```diff
@@ -94,16 +117,26 @@ class BusinessService implements IService {
...
-    payload: Pick<BusinessEntityT, 'companyName'>;
+    payload: BusinessUpdateRequestDto;
+    owner: UserEntityObjectWithGroupT;
   }): Promise<BusinessUpdateResponseDto> {
-    const foundBusinessById = await this.findById(id);
+    const foundBusiness = await this.findByOwnerId(owner.id);
 
-    if (!foundBusinessById) {
+    if (!foundBusiness) {
       throw new NotFoundError({});
```
denred, line 140: status code too

backend/src/packages/trucks/libs/types/types.ts
```diff
@@ -1,3 +1,3 @@
 export { type DriverHaveAccessToTruck } from './driver-have-acces-to-truck.type.js';
 export { type TruckDatabaseModel } from './truck-entity-type/truck-database-model.type.js';
-export { type TruckEntity } from 'shared/build/index.js';
+export { type TruckEntityT } from 'shared/build/index.js';
```
denred: Really, why did you change it?

denred (changes requested):

backend/src/packages/business/business.service.ts
mV22110702: Because for every entity type we prepended `T` to the end, so TruckEntity should also align with these namings.

kalinkaaaa14 (dismissed):

MaxHopaytsa (approved):

PR #274, opened 2023-09-15
th-239: Business Journey: Adding a new driver with generated password on be

denred:
- Removed the password field from the AddFormDriver.
- Added a password generation method.
- Added a method for sending an email with credentials in the form of a login, password, and link.

example


h0wter (approved):

shared/src/packages/drivers/libs/validation-schemas/driver-create-update-request-body.validation-schema.ts
```diff
@@ -6,19 +6,21 @@ import { UserValidationRule } from '~/packages/users/libs/validation-schemas/enu
...
-    .regex(UserValidationRule.DRIVER_LICENSE_NUMBER)
-    .required()
-    .messages({
-      'string.empty': DriverValidationMessage.DRIVER_LICENSE_NUMBER_REQUIRED,
-    }),
+  driverLicenseNumber,
 });
 
-export { driverCreateUpdateRequestBody };
+export { driverCreateUpdateRequestBody, driverLicenseNumber };
```
MaxHopaytsa: Perhaps now it makes sense to rename this validation scheme to driverUpdateRequestBody. After all, it is no longer relevant for creating a driver, and the same applies to the DriverCreateUpdateRequestDto type, rename it to DriverUpdateRequestDto.

backend/src/packages/business/business.controller.ts
```diff
@@ -14,6 +14,7 @@ import {
   type DriverUpdateDeleteRequestParameters,
 } from '../drivers/drivers.js';
 import {
+  driverCreateRequestBody,
   driverCreateUpdateRequestBody,
```
mV22110702: It seems like DriverCreateUpdateRequestDTO is used only in driver's PUT operation, so, it's better to name it DriverUpdateRequestDTO.

backend/src/packages/drivers/driver.service.ts
```diff
@@ -139,6 +146,63 @@ class DriverService implements IService {
...
+<p>${MailContent.CREDENTIALS}</p>
+<p>${MailContent.USERNAME} ${email}</p>
+<p>${MailContent.PASSWORD} ${password}</p><br>
+<p>${MailContent.SIGNIN_LINK}</p>
+<p>${MailContent.LINK}</p><br>
+<p>${MailContent.CLOSING}</p>`,
+    };
+
+    await mailer.send(options);
+  }
```
mV22110702: We have mailService, where you should use your own created View, as it was with PlainView.

shared/src/libs/packages/http/libs/enums/http-error-message.enum.ts
```diff
@@ -22,6 +22,7 @@ const HttpMessage = {
   SHIFT_ALREADY_CLOSED: 'Specified shift is already closed',
   NOT_ACCESS: 'You do not have enough rights to operate this shift',
   ORDER_DOES_NOT_EXIST: 'Order does not exist!',
+  PASSWORD_HASHING: 'Failed to generate hashed password',
```
mV22110702: Seems like PASSWORD_HASHING hasn't been used anywhere.

mV22110702 (changes requested):

backend/src/packages/drivers/driver.service.ts
```diff
@@ -136,9 +140,59 @@ class DriverService implements IService {
...
+    const hasLetter = /[A-Za-z]/.test(password);
+    const hasSpaces = /^\s+$/.test(password);
+
+    if (!hasDigit || !hasLetter || hasSpaces) {
+      return await this.generatePassword();
+    }
+
+    return password;
+  }
+
```
albsbz, line 227: maybe we can use bcrypt for password generaing and then just cut it to length we need

denred: Sure, it's possible, but I think it's not a very good idea. At least, we should need to remove characters and we should check the password length and containing uppercase and lowcase letters and numbers. Yes, it's possible to simplify the generation method by using a string of characters and selecting from it.

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,13 +153,67 @@ class DriverService implements IService {
         userId: user.id,
       }),
     );
-    await this.truckService.addTrucksToDriver(user.id, truckIds);
 
     const driverObject = driver.toObject();
 
+    const emailData: DriverCredentialsViewRenderParameter = {
+      name: `${firstName} ${lastName}`,
```
PavloSukhinin: I believe we have a helper for this

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,13 +153,67 @@ class DriverService implements IService {
...
+
     return { ...user, ...driverObject, group, possibleTruckIds: truckIds };
   }
 
+  private async generatePassword(): Promise<string> {
+    const minLength = 6;
+    const maxLength = 20;
+    const uppercaseChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+    const lowercaseChars = 'abcdefghijklmnopqrstuvwxyz';
+    const numberChars = '0123456789';
```
PavloSukhinin: const UPPER_CASE_BECAUSE_ITS_CONSTANT =

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,13 +153,67 @@ class DriverService implements IService {
...
+
+    const passwordLength =
+      Math.floor(Math.random() * (maxLength - minLength + 1)) + minLength;
```
PavloSukhinin: create a helper for this operation

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,13 +153,67 @@ class DriverService implements IService {
...
+    let password = '';
+
+    password += uppercaseChars.charAt(
+      Math.floor(Math.random() * uppercaseChars.length),
+    );
+    password += lowercaseChars.charAt(
+      Math.floor(Math.random() * lowercaseChars.length),
+    );
+    password += numberChars.charAt(
+      Math.floor(Math.random() * numberChars.length),
```
PavloSukhinin: `Math.floor(Math.random() * something)` this looks like a helper

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,13 +153,67 @@ class DriverService implements IService {
...
+    );
+
+    for (let index = 3; index < passwordLength; index++) {
```
PavloSukhinin: avoid magic numbers (3)

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,13 +153,67 @@ class DriverService implements IService {
...
+      const characterSet = uppercaseChars + lowercaseChars + numberChars;
+      const randomCharacter = characterSet.charAt(
+        Math.floor(Math.random() * characterSet.length),
```
PavloSukhinin: same as above

backend/src/libs/packages/mailer/libs/enums/mail-content.enum.ts
```diff
@@ -0,0 +1,6 @@
+const MailContent = {
+  SUBJECT: 'Credentials for signing in to TowHub app',
+  LINK: 'http://localhost:3000/sign-in',
```
PavloSukhinin: should be configurable or dynamic

PavloSukhinin (changes requested):

shared/src/packages/drivers/libs/types/request/driver-request-create-update-dto.type.ts
```diff
@@ -2,7 +2,7 @@ import { type DriverDto } from '../types.js';
 
 type DriverCreateUpdateRequestDto = DriverDto & {
   password: string;
```
kalinkaaaa14: this dto should not have password inside, remove it
also file naming should be the same as type name

frontend/src/packages/drivers/drivers-api.ts
```diff
@@ -37,7 +37,7 @@ class DriverApi extends HttpApi {
 
   public async addDriver({
     payload,
-  }: DriverAddPayload): Promise<DriverAddResponseWithGroup> {
+  }: Omit<DriverAddPayload, 'reference'>): Promise<DriverAddResponseWithGroup> {
```
kalinkaaaa14: it's strange to add reference to DriverAddPayload type, as it' not in payload, this data is getting on be

backend/src/packages/mail/libs/views/plain/libs/types/plain-view-render-parameter.type.ts
```diff
@@ -1,3 +1,3 @@
-type PlainViewRenderParameter = { body: string };
+type PlainViewRenderParameter = { body?: string };
```
kalinkaaaa14: why body is optional?

backend/src/packages/mail/libs/views/driver-credentials/libs/layout/driver-credentials.hbs
```diff
@@ -0,0 +1,17 @@
...
+    <p>Your credentials for our app are as follows:</p>
+    <ul>
+      <li>Email: {{email}}</li>
+      <li>Password: {{password}}</li>
+    </ul>
+    <p>To sign in, please click the following link:</p>
+    <a href='{{signInLink}}'>Sign In</a>
+    <p>Thank you for using our app!</p>
+  </body>
+</html>
```
kalinkaaaa14, line 17: backend/src/packages/mail/libs/views/plain/libs/layout/driver-credentials.hbs
duplicates?

backend/src/packages/drivers/libs/helpers/get-random-character.helper.ts
```diff
@@ -0,0 +1,4 @@
+const getRandomCharacter = (chars: string): string =>
+  chars.charAt(Math.floor(Math.random() * chars.length));
```
kalinkaaaa14: curly + return

backend/src/packages/drivers/libs/helpers/get-password-length.helper.ts
```diff
@@ -0,0 +1,4 @@
+const getPasswordLength = (maxLength: number, minLength: number): number =>
+  Math.floor(Math.random() * (maxLength - minLength + 1)) + minLength;
```
kalinkaaaa14: curly + return

backend/src/packages/drivers/driver.service.ts
Some14u: Yes, in shared

backend/src/packages/drivers/driver.service.ts
```diff
@@ -154,11 +161,65 @@ class DriverService implements IService {
...
+    const MIN_LENGTH = 6;
+    const MAX_LENGTH = 20;
+    const UPPER_CASE_CHARS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+    const LOWER_CASE_CHARS = 'abcdefghijklmnopqrstuvwxyz';
+    const NUMBER_CHARS = '0123456789';
+
+    const passwordLength = getPasswordLength(MIN_LENGTH, MAX_LENGTH);
+
+    const charSets = [LOWER_CASE_CHARS, UPPER_CASE_CHARS, NUMBER_CHARS].sort(
+      () => Math.random() - 0.5,
```
kalinkaaaa14: move out this magic num

shared/src/packages/drivers/libs/types/request/driver-request-create-update-dto.type.ts
denred: If the driver decides to update their password, does that mean we'll have to create another data type?

backend/src/packages/mail/libs/views/driver-credentials/libs/layout/driver-credentials.hbs
denred: It's not a duplicate, it's another template. The base template requires a body, which is very generic. I think it's better to use a template that requires variables for insertion.

backend/src/packages/mail/libs/views/plain/libs/types/plain-view-render-parameter.type.ts
denred: Because in the current implementation, we can't use other variables.

shared/src/packages/drivers/libs/types/request/driver-create-request-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type DriverDto } from '../types.js';
+
+type DriverCreateRequestDto = DriverDto & {
+  truckIds?: number[];
```
kalinkaaaa14: why it's optional? can we create a driver without trucks?

backend/src/libs/packages/controller/libs/types/default-api-handler-options.type.ts
```diff
@@ -6,6 +6,7 @@ type DefaultApiHandlerOptions = {
   params?: unknown;
   user?: unknown;
   parsedFiles?: MultipartParsedFile[];
+  hostname?: unknown;
```
kalinkaaaa14: why unknown here?

kalinkaaaa14 (dismissed):

MaxHopaytsa (approved):

h0wter (approved):

mV22110702 (approved):

kalinkaaaa14 (approved):

PR #282, opened 2023-09-16
th-281: Fixed view types

mV22110702:


denred (approved):

AlexVlaso (approved):

backend/src/packages/mail/libs/maps/template-name-to-view.map.ts
```diff
@@ -1,11 +1,8 @@
-import { type IView } from '~/libs/packages/packages.js';
-
 import { TemplateName } from '../enums/enums.js';
-import { type TemplateNameValues } from '../types/types.js';
 import { plainView } from '../views/views.js';
 
-const templateNameToView: Record<TemplateNameValues, IView> = {
+const templateNameToView = {
```
kalinkaaaa14: add a type for this map

mV22110702: Why should there be a type for it, if it's inferred from the actual values?

mV22110702: > add a type for this map
Why should there be a type for it, if it's inferred from the actual values?

mV22110702: > add a type for this map
Why should there be a type if it's inferred and works by the same principle as enum type?

PavloSukhinin (dismissed):

backend/src/packages/mail/libs/maps/template-name-to-view.map.ts
kalinkaaaa14: we need to add type here, so we can't assign any other value as one of values from TemplateNameValues to key, and no other to it's value as IView
now I can write whatever I want as a key, and ts won't tell me that something was wrong

mV22110702: Added type, but I had to explicitly specify its keys and values to keep render parameter type.

kalinkaaaa14: @mV22110702 Record<TemplateNameValues, IView> won't it be suitable?

mV22110702: IView is generic, so, unfortunately, it's not going to work.

PR #285, opened 2023-09-16
th-241: Bug: With an active session in two browsers, when reloading the page on one of them, the user's session ends and a 404 error appears

denred:
- Fixed access token generation


h0wter (approved):

backend/src/packages/auth/auth.service.ts
```diff
@@ -235,6 +239,28 @@ class AuthService {
...
+    }
+
+    const currentTimestamp = Date.now();
+    const tokenIssuedAtTimestamp = payload.iat * 1000;
+    const timeDifferenceInMilliseconds = Math.abs(
+      currentTimestamp - tokenIssuedAtTimestamp,
+    );
+
+    const lifetimeInMilliseconds =
+      Number.parseInt(accessLifetime) * 60 * 60 * 1000;
```
albsbz: Better is to replace magic numbers with variables

AlexVlaso (approved):

PavloSukhinin (approved):

PR #293, opened 2023-09-18
th-263: Log out implementation

h0wter:
Added the ability to log out a user with token deletion from local storage and database.
Changed the `li` component in the burger menu list, when a picture was used the click on the element triggered only on a small area of the picture, now the click triggers on the entire area of the element.


h0wter:
Removed the redirect as some pages are accessible to unregistered users.

frontend/src/packages/auth/auth-api.ts
```diff
@@ -77,6 +77,14 @@ class AuthApi extends HttpApi {
       CustomerSignUpResponseDto | BusinessSignUpResponseDto
     >();
   }
+
+  public async logOut(): Promise<unknown> {
+    return await this.load(this.getFullEndpoint(AuthApiPath.LOGOUT, {}), {
+      method: 'GET',
+      contentType: ContentType.JSON,
+      hasAuth: true,
```
albsbz, line 85: will it work if token expired?

h0wter: I have connected a JWT validation strategy on endpoint, and it checks for token relevance. With an expired one, the user will not be able to logout.

h0wter:
Fixed the background for the password hide/display icon to be transparent during autocomplete.

kalinkaaaa14 (approved):

Druxight (approved):

PR #298, opened 2023-09-20
th-280: Bug: When trying to create a shift for the driver, the 404 error (in Postman) and a 500 error (in swagger) appear for an unauthorized user

TetianaDoroshko:
Fixed authorization checking.
Fixed path in swagger.


backend/src/packages/shifts/shift.controller.ts
```diff
@@ -39,7 +39,7 @@ class ShiftController extends Controller {
     this.addRoute({
       path: ShiftsApiPath.ROOT,
       method: 'POST',
-      authStrategy: AuthStrategy.INJECT_USER,
+      authStrategy: [AuthStrategy.VERIFY_JWT, AuthStrategy.INJECT_USER],
       validation: {
```
albsbz, line 43: It seems that difference between AuthStrategy.VERIFY_JWT and AuthStrategy.INJECT_USER that first throws error on unauthorized user and the second one not, so we need to use only one of them

TetianaDoroshko: It turns out that INJECT_USER only attach user to request. And if there isn't token, it attaches user=undefined and allowed to get to the private endpoints.

albsbz: Then we can use only AuthStrategy.VERIFY_JWT

TetianaDoroshko: Yes, you are right. Thank you.

kalinkaaaa14 (approved):

Issue #302, opened 2023-09-20
Bug: The Driver License Number Field in the Add Driver modal accepts only Ukrainian license numbers as valid


OlenaCherneta:
**Preconditions:**
 User is logged in as business

**Steps to reproduce:**

1. Open the Dashboard/Drivers.
2. Click on the Add driver button
3. Fill the Driver License Number Field with valid non-Ukrainian license number, for example [this](https://www.licencia-conducir.com/wp-content/uploads/%C2%BFWhat-is-my-drivers-license-number-1.jpg)
4. Click on the Add Driver button
5. Pay attention to the validation warning

**Actual result**
The non-Ukrainian driver license number is rejected as invalid. Non-user friendly validation warning is shown.

**Expected result**
The non-Ukrainian driver license number successfully added, no warnings appeared. In case of invalid value the user-friendly validation warning is displayed

**Environment:**
Windows 10 Pro x64, Google Chrome v. 117.0.5938.89

**Attachment (see below)**

https://github.com/BinaryStudioAcademy/bsa-2023-towhub/assets/75045481/5d274259-90f5-435d-bf4b-2716f45fe204

[Validation rules/warnings](https://docs.google.com/document/d/17MGV3liHVc2qyIjLQJTFgQcNd1WNajcQFcp0fTwDy7Q/edit?usp=sharing
):
Input:
accept uppercase Latin letters;
accept digits;
accept space and hyphen (-);
minimal length - 5;
maximum length - 30;
should not contain only spaces;
required field;
Validation Messages:
For no content in the field: "This field is mandatory"
For the non-valid value: “This driver license number does not seem valid”

Some14u:
resolved #306

Some14u:
fixes #306

Some14u:
Link #306

Some14u:
Resolve #306

Some14u:
Closes #306

PR #306, opened 2023-09-20
th-302: Make Proper Driver License Number Validation

Some14u:
[Bug ticket](https://github.com/BinaryStudioAcademy/bsa-2023-towhub/issues/302)
I did the rules in accordance to the requirements.
изображение
It counts only significant characters while testing the length constraints:
изображение


Some14u:
Resolves #302

shared/src/packages/drivers/libs/enums/driver-validation-rule.enum.ts
```diff
@@ -0,0 +1,8 @@
+const DriverValidationRule = {
+  LICENSE_MIN_LENGTH: 5,
+  LICENSE_MAX_LENGTH: 30,
+  LICENSE_NUMBER: /^[\dA-Z]+(?:[ -][\dA-Z]+)*$/,
```
albsbz, line 4:
```suggestion
  LICENSE_NUMBER: /^(?=.{5,30}$)[\dA-Z]+(?:[ -][\dA-Z]+)*$/,
```

Some14u: This would count all symbols. I'm pretty sure, that min 5 max 30 constraint means only meaningful symbols, like letters and numbers. This does not include hyphens (-) and whitespaces. Otherwise what is the point of restriction, if the same number can be written as 1-2-3-4-5(9 characters, but 5 meaningful symbols) or 12345 (5 characters and 5 meaningful symbols).

I'm gonna leave it like this by now unless somebody prove I'm wrong.

h0wter (approved):

AlexVlaso (approved):

shared/src/packages/drivers/libs/enums/driver-validation-rule.enum.ts
PavloSukhinin: What will be stored in the database? Value with spaces or without them? I'm not sure we want to store things like this `FOO1 2 3 45 A A A A A A A A A A A A A A A A A A A A A A`

denred: It looks strange to use regex in an enum that has fields with string length constraints. You could include it directly in the regex.

Some14u: So what are you suggestions? there is no defined template for license number. All we know, that it is possible to have numbers, letters, hyphens and spaces. The template i use checks exactly this, with constraint that you cannot use combination of double hyphen/spaces.

Speaking on checking the length, though, it is restricted to 5...30 MEANINGFUL characters (letters and symbols).

If you want me to prohibit hyphens and spaces, i'm okay about it. If you want me to count hyphens and spaces as meaningful letters, i'm not fan of that solution, because this makes no sense to restrict the value that way to me.

Some14u: @PavloSukhinin there are certain formats of writing license plates, they may or may not include hyphens and spaces. 
If you want it so, i can strip all those delimeters and store a pure meaningful value in the db. But it will lack proper definition. Not sure, if it is okay to have AB123456 instead of AB 123-456 (if the country respects this precise format)

Some14u: >It looks strange to use regex in an enum that has fields with string length constraints. You could include it directly in the regex.

@denred i'm quite opposite to this

do  you think it is okay to hide the length constraints in something like this : 
```
/^(?=.{5,30}$)[\dA-Z]+(?:[ -][\dA-Z]+)*$/,
(find the limits in 10 seconds :) )

```
vs this

```
  LICENSE_MIN_LENGTH: 5,
  LICENSE_MAX_LENGTH: 30,
  LICENSE_NUMBER: /^[\dA-Z]+(?:[ -][\dA-Z]+)*$/,
```

denred: I meant that it would be a good idea to move the regex into a separate file or variable, rather than keeping it in the enum. It would also be good to specify the limits of the license number directly in the regex and extract the length parameters into an enum as you already have.

PavloSukhinin (approved):

PR #325, opened 2023-09-22
th-286: Edit form for business and customer users

MaxHopaytsa:
- Create edit form
- Add redux part
- Add backend part
- Add swager
edit


frontend/src/pages/profile/components/edit-form/edit-form.tsx
```diff
@@ -0,0 +1,63 @@
...
+            phone,
+            email,
+            taxNumber: business.taxNumber,
+            companyName: business.companyName,
+          }}
+          validationSchema={businessEditValidationSchema}
+          onSubmit={onSubmit}
+          btnLabel="Save"
+          fields={editBusinessFields}
+        />
```
denred: It looks like a switch statement.

frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const handleSubmit = useCallback(
+    (payload: CustomerEditDto | BusinessEditDto): void => {
+      if (mode === AuthMode.CUSTOMER) {
+        void dispatch(authActions.editCustomer(payload));
+      }
+
+      if (mode === AuthMode.BUSINESS) {
+        void dispatch(authActions.editBusiness(payload as BusinessEditDto));
+      }
```
denred: It looks like a switch statement, too

denred (changes requested):

frontend/src/pages/profile/components/edit-form/edit-form.tsx
```diff
@@ -0,0 +1,63 @@
+import { type UserEntityObjectWithGroupAndBusinessT } from 'shared/build';
```
MaxHopaytsa: I'll fix it

backend/src/packages/users/user.controller.ts
```diff
@@ -45,6 +72,22 @@ class UserController extends Controller {
           }>,
         ),
     });
+
+    this.addRoute({
+      path: UsersApiPath.ROOT,
+      method: 'PUT',
+      authStrategy: defaultStrategies,
```
h0wter: You can pass the defaultStrategies to the super constructor and they don't have to be specified on the route. Only if you want to override or append.  `super(logger, ApiPath.USERS, defaultStrategies);`
In that case, you could leave only `AuthStrategy.VERIFY_JWT` in the `defaultStrategies`, remove the strategies from the GET route, and make `authStrategy: [...defaultStrategies, AuthStrategy.VERIFY_CUSTOMER_GROUP]` in the PUT route.

frontend/src/packages/users/libs/enums/user-successful-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const UserNotificationMessage = {
+  SUCCESS_EDIT_USER: 'You update your personal data!',
```
mV22110702:
```suggestion
  SUCCESS_EDIT_USER: 'You have updated your personal data!',
```

shared/src/packages/users/libs/validation-schemas/business-edit.validation-schema.ts
```diff
@@ -0,0 +1,35 @@
...
+    }),
+  taxNumber: joi
+    .string()
+    .trim()
+    .required()
+    .pattern(createPatternByRule(Rule.TAX_NUMBER))
+    .messages({
+      'string.empty': Message.FIELD_IS_REQUIRED,
+      'string.pattern.base': Message.TAX_NUMBER_NOT_VALID,
+    }),
```
mV22110702: It would be great if we create reusable joi fields for companyName and taxNumber, because we defined them earlier in schema for business creation.

mV22110702 (changes requested):

mV22110702:
Also, DELETE and GET business endpoints (and corresponding business service methods) should also not rely on id, but there's already PR on that fix, so we can keep them as is in this PR.

frontend/src/libs/enums/app-route.enum.ts
```diff
@@ -6,7 +6,8 @@ const AppRoute = {
   SIGN_UP_BUSINESS: '/sign-up/business',
   SIGN_UP_CUSTOMER: '/sign-up/customer',
   ORDER_HISTORY: '/order-history',
-  EDIT_PROFILE: '/edit-profile',
+  EDIT_CUSTOMER_PROFILE: '/edit-customer-profile',
+  EDIT_BUSINESS_PROFILE: '/edit-business-profile',
```
h0wter: How good an idea is it to do an endpoint of `/edit-...-profile`? Would it be better to just do `/profile`? I think this option is easier to extend, if you need to add a page for a profile, then editing can take place right there, and the user doesn't need to be redirected to another route to do it.

And it seems to me to use the option with one endpoint for both users, and in the component check the group and render the necessary component depending on the group is better. When the user will log in and see `/edit-business-profile` he may think what other profile he has? This is my opinion, not sure if it is correct.

frontend/src/packages/business/business-api.ts
```diff
@@ -49,6 +53,22 @@ class BusinessApi extends HttpApi {
 
     return await response.json<TruckEntity>();
   }
+
+  public async editBusiness(
+    payload: BusinessEditDto,
```
h0wter, line 82: The type is named `BusinessEdit`, the method is `editBusiness`, maybe make the same, for example `EditBusinessDto`?

frontend/src/pages/profile/components/edit-form/edit-form.tsx
```diff
@@ -0,0 +1,85 @@
...
+  onSubmit: (payload: CustomerEditDto | BusinessEditDto) => void;
+  mode: ValueOf<typeof AuthMode>;
+};
+
+const EditForm: React.FC<Properties> = ({ onSubmit, mode }: Properties) => {
+  const { firstName, lastName, phone, email, business } = useAppSelector(
+    selectUser,
+  ) as UserEntityObjectWithGroupAndBusinessT;
+
+  const getScreen = useCallback((): React.ReactNode => {
```
h0wter, line 33: Maybe put it in the hook to make the component cleaner?

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -36,51 +44,80 @@ const { reducer, actions, name } = createSlice({
...
+          signIn.pending,
+          getCurrent.pending,
+          logOut.pending,
+        ),
+        (state) => {
+          state.dataStatus = DataStatus.PENDING;
+        },
+      );
+    builder.addMatcher(
+      isAllOf(signUp.rejected, signIn.rejected, logOut.rejected),
```
h0wter:
```suggestion
      isAnyOf(signUp.rejected, signIn.rejected, logOut.rejected),
```

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -36,51 +44,80 @@ const { reducer, actions, name } = createSlice({
...
+          };
+        }
+        state.dataStatus = DataStatus.FULFILLED;
+      })
+      .addCase(logOut.fulfilled, (state) => {
+        state.user = initialState.user;
+        state.dataStatus = DataStatus.FULFILLED;
+      })
+      .addMatcher(
+        isAllOf(
```
h0wter: I think there should be an isAnyOf here.
```suggestion
        isAnyOf(
```

frontend/src/slices/auth/auth.slice.ts
MaxHopaytsa: Yes, it was an accident, thanks for noticing

frontend/src/packages/business/business-api.ts
MaxHopaytsa: In our project, such types start with the name of the entity. And all methods, of course, start with the verb

frontend/src/slices/auth/auth.slice.ts
MaxHopaytsa: The same)

frontend/src/packages/business/business-api.ts
h0wter: Got it, thanks.

frontend/src/pages/profile/components/edit-form/edit-form.tsx
MaxHopaytsa: This can certainly be done, but I don't think it's necessary. This code is not reusable, it is used only in this component, and it does not overload it with code. So it's 50/50, but I would leave it as it is for now.

h0wter (approved):

mV22110702 (approved):

Some14u (approved):

frontend/src/slices/auth/actions.ts
```diff
@@ -42,6 +46,50 @@ const signUp = createAsyncThunk<
...
 
+const editCustomer = createAsyncThunk<
+  CustomerEditDto,
+  CustomerEditDto,
+  AsyncThunkConfig
+>(`${sliceName}/edit-customer`, async (payload, { extra, rejectWithValue }) => {
+  const { userApi, notification } = extra;
+
+  try {
+    const editCustomer = await userApi.editCustomer(payload);
```
PavloSukhinin: `editedCustomer =`?

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -44,61 +47,87 @@ const { reducer, actions, name } = createSlice({
...
+      })
+      .addCase(signUp.fulfilled, (state, action) => {
+        state.error = null;
+        state.user = action.payload;
+        state.dataStatus = DataStatus.FULFILLED;
+      })
+      .addCase(editBusiness.fulfilled, (state, action) => {
+        const { firstName, lastName, phone, email, business } = action.payload;
+
+        if (state.user) {
```
kalinkaaaa14: is there any logic in it? I think edit business can be fullfilled only for authorized user, so user should be present in state always

frontend/src/packages/users/libs/enums/user-successful-message.enum.ts
kalinkaaaa14, line 1: file naming should be same as enum naming

frontend/src/pages/profile/components/edit-form/edit-form.tsx
```diff
@@ -0,0 +1,72 @@
...
+            />
+          );
+        }
+      }
+    }
+  }, [business, email, firstName, group.key, lastName, onSubmit, phone]);
+
+  return (
+    <div className={styles.formWrapper}>
+      <h3 className={getValidClassNames('h4', 'uppercase', styles.title)}>
```
kalinkaaaa14: why tag is h3 and classname is 'h4' 😃

frontend/src/libs/components/protected-route-for-business-customer/protected-route-for-business-customer.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const isLoading = useAppSelector(selectIsLoading);
+  const user = useAuthUser();
+
+  if (isLoading === DataStatus.PENDING) {
+    return <Spinner size="sm" />;
+  }
+
+  return user &&
+    (user.group.key === UserGroupKey.CUSTOMER ||
+      user.group.key === UserGroupKey.BUSINESS) ? (
```
kalinkaaaa14: let's move it to the separate variable and remove check for user here, it's redundant

```
if(isBusinessOrCustomer) { return <RouterOutlet /> }

return <NotFound />
```

kalinkaaaa14 (changes requested):

kalinkaaaa14 (dismissed):

frontend/src/slices/auth/actions.ts
```diff
@@ -51,6 +55,50 @@ const signUp = createAsyncThunk<
...
+  try {
+    const editedCustomer = await userApi.editCustomer(payload);
+
+    notification.success(UserNotificationMessage.SUCCESS_EDIT_USER);
+
+    return editedCustomer;
+  } catch (error_: unknown) {
+    const error = error_ as HttpError;
+
+    notification.error(getErrorMessage(error.message));
```
denred: It would be nice to see an error message below the specific field in the form of a parameter is invalid so that we don't have to close the form and re-enter the same data. Notifications should be provided only for system errors or errors in the completed action.

frontend/src/slices/auth/actions.ts
```diff
@@ -51,6 +55,50 @@ const signUp = createAsyncThunk<
...
+  BusinessEditResponseDto,
+  BusinessEditDto,
+  AsyncThunkConfig<HttpError>
+>(`${sliceName}/edit-business`, async (payload, { extra, rejectWithValue }) => {
+  const { businessApi, notification } = extra;
+
+  try {
+    const editedBusiness = await businessApi.editBusiness(payload);
+
+    notification.success(UserNotificationMessage.SUCCESS_EDIT_USER);
```
denred, line 88: the same

denred (changes requested):

kalinkaaaa14 (approved):

PR #331, opened 2023-09-22
th-319: Fix duplicated truck after create

MaxHopaytsa:
- Fix truck slice
 - Made the table more responsive for different screen sizes
 - Fix truck capacity


h0wter (approved):

AlexVlaso (approved):

frontend/src/slices/trucks/trucks.slice.ts
```diff
@@ -90,7 +90,16 @@ const { reducer, actions, name } = createSlice({
   extraReducers(builder) {
     builder
       .addCase(addTruck.fulfilled, (state, action) => {
-        state.trucks.unshift(action.payload);
+        const newTruck = action.payload;
+
+        const isTruckUnique = !state.trucks.some(
+          (truck) => truck.id === newTruck.id,
+        );
```
kalinkaaaa14, line 97: I don't understand this check here. 
Am I right that on be we can create new duplicated truck, and on fe we are checking it after successfull saving on be? 🤔

MaxHopaytsa: No, there is no duplication on be. There is duplication only on the UI. I use **unshift** so that I can add a truck to the top of the table on the page where the user is now. However, there may be such a situation that the page has been refreshed and due to sorting this added truck is already on this page, and we are on the UI, and using **unshift** we add it to the top of the table, and it looks like we have two identical trucks on one page, but this is only on the UI. When the page is reloaded, this additional truck will disappear. That's why I added this check to avoid such a case. If the page has already received a truck, we do not add it to the top, and if the truck is to be added to another page on be, and not to the current one, we temporarily move it to the top on fe so that the user can see that he has created a truck.

shared/src/packages/trucks/libs/enums/truck-capacity.enum.ts
```diff
@@ -1,6 +1,6 @@
 const TruckCapacity = {
   MIN: 1,
-  MAX: 4,
+  MAX: 999,
```
PavloSukhinin, line 3: why? 🥲

MaxHopaytsa: The QA team said that the capacity should be 1-999, and they even created a separate bug ticket for this. I can revert it back, it will take a minute))

PavloSukhinin: If QAs said so... 🥲

PavloSukhinin (approved):

PR #346, opened 2023-09-23
th-337: Remove time field from order form

DmytroVoronkov:
- removed time field
- fixed form size
- added scroll to form
- set 50/50 ratio to page
- fixed validation errors

image


h0wter (approved):

frontend/src/pages/order/libs/components/styles.module.scss
```diff
@@ -1,28 +1,14 @@
 @import "src/assets/css/vars.scss";
 
 .formWrapper {
+  flex-grow: 1;
   width: 100%;
-  max-width: 700px;
-  height: calc(100vh - 271px);
-  padding: 20px 100px 35px;
+  padding: 20px 15% 35px;
```
kalinkaaaa14: can we write padding in px?

frontend/src/pages/order/libs/components/libs/helpers.ts
```diff
@@ -2,13 +2,19 @@ import { type OrderCreateRequestDto } from '~/packages/orders/orders.js';
 
 import { TruckCarsQuantity } from './enums.js';
 
+const getCurrentDate = (): string => {
+  const currentDatetime = new Date();
+
+  return currentDatetime.toISOString().slice(0, 16);
```
kalinkaaaa14: why do you need to use slice? 
add a separate helper for it

kalinkaaaa14:
as we decided that we will be considering all orders = asap orders then actually scheduled time is not needed at our db. 
but it the future we can implement planning future orders, so ok, let's leave this column in order table

**but scheduled time it's better to calculate on be, so remove it from the default values on fe, and add to service layer calculation of scheduled time field.**

h0wter (approved):

shared/src/packages/orders/libs/enums/orders-validation-message.enum.ts
```diff
@@ -5,6 +5,9 @@ const OrdersValidationMessage = {
   DRIVER_ID_MUST_BE_NUMBER: 'Orders driverId must be number',
   TRUCK_ID_MUST_BE_NUMBER: 'Orders truckId must be number',
   CARS_QTY_MUST_BE_NUMBER: 'Orders carsQty must be number',
+  FIELD_IS_REQUIRED: 'This field is mandatory',
+  NAME_NOT_VALID: 'Must be 1 - 40 characters, start with a letter, no digits',
+  PHONE_NOT_VALID: 'Must be 7-18 digits, start with +',
```
Druxight: Same messages but different range format. 
```suggestion
  NAME_NOT_VALID: 'Must be 1 - 40 characters, start with a letter, no digits',
  PHONE_NOT_VALID: 'Must be 7 - 18 digits, start with +',
```

Druxight (changes requested):

AlexVlaso (approved):

denred (approved):

PR #347, opened 2023-09-23
th-336: When creating an order, the phone and username values are not stored in the database

AlexVlaso:


backend/src/packages/orders/order.controller.ts
```diff
@@ -410,11 +415,16 @@ class OrderController extends Controller {
...
+    const customerPhone = options.body.customerPhone ?? options.user?.phone;
+    const customerName = options.body.customerName ?? options.user?.firstName;
+
     return {
       status: HttpCode.OK,
       payload: await this.orderService.create({
         ...options.body,
         userId: options.user?.id ?? null,
+        customerPhone: customerPhone ?? null,
+        customerName: customerName ?? null,
```
h0wter: I think it would be a little cleaner that way.
```suggestion
    const userId = options.user?.id ?? null;
    const customerPhone = options.body.customerPhone ?? options.user?.phone ?? null;
    const customerName = options.body.customerName ?? options.user?.firstName ?? null;

    return {
      status: HttpCode.OK,
      payload: await this.orderService.create({
        ...options.body,
        userId,
        customerPhone,
        customerName,
```

backend/src/packages/orders/order.controller.ts
```diff
@@ -410,11 +415,19 @@ class OrderController extends Controller {
       user: UserEntityObjectWithGroupT | null;
     }>,
   ): Promise<ApiHandlerResponse<OrderResponseDto>> {
+    const userId = options.user?.id ?? null;
+    const customerPhone =
+      options.body.customerPhone ?? options.user?.phone ?? null;
+    const customerName =
+      options.body.customerName ?? options.user?.firstName ?? null;
```
kalinkaaaa14: contollers should be clean. pass user to service and then decide what to save into db

rewrite to:
const { firstName = null, phone = null } = user ?? {};
const nameInOrder = customerName ?? firstName;
const phoneInOrder = customerPhone ?? phone;

h0wter (approved):

kalinkaaaa14 (approved):

PR #358, opened 2023-09-24
th-351: Sidebar for drivers

TetianaDoroshko:
after authorization, the driver should see a page with a sidebar
after sign-in driver is redirected to choosing Truck (/available-trucks)
when driver choose truck, he will be redirected to Orders page (Trucks is not available)
if shift is open, sidebar contain a End Shift button. When it is pressed, shift will be closed and driver redirects to Truck.
image


h0wter (approved):

DmytroVoronkov (approved):

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
```diff
@@ -0,0 +1,43 @@
+import { Button } from '~/libs/components/components.js';
+import { getValidClassNames } from '~/libs/helpers/helpers.js';
+import {
+  useAppDispatch,
+  useAppSelector,
+  useCallback,
+} from '~/libs/hooks/hooks.js';
+import { actions as driverActions } from '~/slices/driver/driver.js';
+import { ShiftStatus } from '~/slices/driver/libs/enums/shift-status.enum';
```
h0wter: It's missing .js and it's better to do a reexport.
```suggestion
import { ShiftStatus } from '~/slices/driver/libs/enums/enums.js';
```

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
h0wter, line 1: I think it is better to create a separate folder for the component.
components/sidebar/components/end-shift-button/end-shift-button.tsx

frontend/src/pages/dashboard/components/sidebar/components/styles.module.scss
h0wter, line 1: Same here.
components/sidebar/components/end-shift-button/styles.module.scss

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </p>
+      <Button
+        label={'End shift'}
+        frontIcon={'truck'}
+        onClick={handleClick}
+        className={styles.button}
+      />
+    </div>
+  ) : null;
+};
```
h0wter:
```suggestion
  return isActive && (
    <div className={styles.block}>
      <p className={getValidClassNames(styles.truckInfo, 'textMd')}>
        {truck?.manufacturer} {truck?.licensePlateNumber}
      </p>
      <Button
        label={'End shift'}
        frontIcon={'truck'}
        onClick={handleClick}
        className={styles.button}
      />
    </div>
  );
};
```
But it seems to me that this button should not check if there is a shift, it is better to do it in the parent component, this logic does not apply to the button, my opinion.
If we want to show the button always, but make it inactive in case of no shift - then we need to use `isDisabled` props for the button. Otherwise we get a component that is rendered but returns nothing if there is no active shift.
And we can pass truck as props, then we don't have to check `shiftStatus === ShiftStatus.ACTIVE` and whether truck has fields.

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const isActive = shiftStatus === ShiftStatus.ACTIVE;
+
+  const handleClick = useCallback(() => {
+    void dispatch(driverActions.endShift());
+  }, [dispatch]);
+
+  return isActive ? (
+    <div className={styles.block}>
+      <p className={getValidClassNames(styles.truckInfo, 'textMd')}>
+        {truck?.manufacturer} {truck?.licensePlateNumber}
```
h0wter: Use `manufacturer-key-to-readable-name.map` to convert the value to the correct name.
```suggestion
        {manufacturerKeyToReadableName[manufacturer]} {truck?.licensePlateNumber}
```

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </p>
+      <Button
+        label={'End shift'}
+        frontIcon={'truck'}
```
h0wter:
```suggestion
        frontIcon={IconName.TRUCK}
```

frontend/src/pages/dashboard/components/sidebar/libs/helpers.ts
```diff
@@ -1,7 +1,7 @@
-import { AppRoute } from '~/libs/enums/app-route.enum';
+// import { AppRoute } from '~/libs/enums/app-route.enum';
```
h0wter: Why is this comment here?

frontend/src/pages/dashboard/components/sidebar/libs/helpers.ts
```diff
@@ -1,7 +1,7 @@
...
 
 const checkActiveTab = (path: string, tab: string): boolean => {
-  return path === `${AppRoute.DASHBOARD}/${tab}`;
+  return path === tab;
```
h0wter: Is it necessary to create a separate helper for this logic?

frontend/src/pages/dashboard/components/sidebar/sidebar.tsx
```diff
@@ -68,6 +69,7 @@ const Sidebar: React.FC<Properties> = ({ isCollapsed = false }: Properties) => {
       )}
     >
       <ul className={styles.list}>{renderTabs()}</ul>
+      {user?.group.key === UserGroupKey.DRIVER && <EndShiftButton />}
```
h0wter: Maybe put this condition in a variable and use it in getTabs too? 
`const isUserDriver = user?.group.key === UserGroupKey.DRIVER;`
```suggestion
      {isUserDriver && <EndShiftButton />}
```

h0wter (changes requested):

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
TetianaDoroshko: About replace ternary to &&. Functional component has to return 'ReactElement<any, any> | null', so I can't use use "&&" and return undefined.
About responsibility of checking the shift. I think Sidebar doesn't have to do that too. It just check the group of the user and render EndSfiftButton component. Then EndSfiftButton check the status of shift and decides to render the button or not. If it returns null the elements won't add to DOM.
Let's wait for coach's opinion.

h0wter: If the component should return something, you can use a react fragment. It adds a bit of code, but it still looks cleaner than with a ternary operator. Just wrap the entire return in `<>...</>`.

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button/end-shift-button.tsx
```diff
@@ -1,12 +1,14 @@
 import { Button } from '~/libs/components/components.js';
+import { IconName } from '~/libs/enums/icon-name.enum';
```
h0wter:
```suggestion
import { IconName } from '~/libs/enums/icon-name.enum.js';
```

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button/end-shift-button.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { getValidClassNames } from '~/libs/helpers/helpers.js';
 import {
   useAppDispatch,
   useAppSelector,
   useCallback,
 } from '~/libs/hooks/hooks.js';
+import { manufacturerKeyToReadableName } from '~/packages/trucks/libs/maps/manufacturer-key-to-readable-name.map';
```
h0wter:
```suggestion
import { manufacturerKeyToReadableName } from '~/packages/trucks/libs/maps/manufacturer-key-to-readable-name.map.js';
```

h0wter (changes requested):

frontend/src/pages/dashboard/components/sidebar/sidebar.tsx
```diff
@@ -68,6 +68,7 @@ const Sidebar: React.FC<Properties> = ({ isCollapsed = false }: Properties) => {
       )}
     >
       <ul className={styles.list}>{renderTabs()}</ul>
+      {isDriver && <EndShiftButton />}
```
kalinkaaaa14, line 72: if the shift will be ended automatically, will this button disappear?

frontend/src/pages/dashboard/components/sidebar/sidebar.tsx
```diff
@@ -39,7 +39,7 @@ const Sidebar: React.FC<Properties> = ({ isCollapsed = false }: Properties) => {
       <li
         className={getValidClassNames(
           styles.item,
-          checkActiveTab(location.pathname, tab.path) && styles.active,
+          location.pathname === tab.path && styles.active,
```
kalinkaaaa14: why do you remove checkIsActiveTab?

frontend/src/pages/dashboard/components/sidebar/components/end-shift-button.tsx
kalinkaaaa14: I think we can leave logic for displaying the button here, in this separate component and not to place it inside sidebar
but also rewrite it with react fragment and && as Vlad suggested

frontend/src/pages/dashboard/components/sidebar/sidebar.tsx
h0wter: @kalinkaaaa14 It's my fault, the previous version of the helper did a `path === `${AppRoute.DASHBOARD}/${tab}`` comparison. Now it's just comparing incoming values without any additional logic over them. It seemed to me that this is a complication and the comparison can be done in place like we have in other parts of the application.

h0wter: @kalinkaaaa14 It's my fault, the previous version of the helper did a `path === ${AppRoute.DASHBOARD}/${tab}` comparison. Now it's just comparing incoming values without any additional logic over them. It seemed to me that this is a complication and the comparison can be done in place like we have in other parts of the application.

TetianaDoroshko: it's ok, I'll change it

h0wter (approved):

kalinkaaaa14 (approved):

PavloSukhinin (approved):

PR #362, opened 2023-09-24
th-357: Quality criteria fixes

AlexVlaso:


backend/src/packages/map/libs/constants/constants.ts
```diff
@@ -1,3 +1,4 @@
 const METERS_IN_ONE_KM = 1000;
+const DECIMAL = 2;
```
Some14u:
```suggestion
const PRECISION = 2;
```

frontend/src/libs/components/file-input/file-input.tsx
```diff
@@ -86,8 +87,10 @@ const FileInput = <T extends FieldValues & FileFormType>({
 
   const handleDrop = useCallback(
     (acceptedFiles: File[], rejectedFiles: FileRejection[]) => {
+      const isNotEmptyRejectedFiles = rejectedFiles.length > ZERO_VALUE;
```
Some14u: @kalinkaaaa14 are you sure this is the right way to do this?
I guess this is a overkill. The whole point of having "speaking names" is to assign a meaningful context to the number.
what do we add by changing 0 to ZERO_VALUE? Is it now more clear what it does? Moreover, such name prevents us form changing the value in the future (not relevant for this particular case, but nevertheless).

I doubt we should change each and every code 0 and 1 to ZERO_VALUE and ONE_VALUE just to be on the safe side about following the rule B31.

I suggest to revert it back to 0 and use a common sense and sanity every time we're following some particular rule.

h0wter:
It's interesting to see what mistakes we've made :)

frontend/src/libs/packages/map/libs/helpers/rotate-img.helper.ts
```diff
@@ -4,16 +4,20 @@ const calculateRad = (rotationAngle: number): number => {
...
   image: HTMLImageElement,
   rotationRad: number,
 ): void => {
   const context = canvas.getContext('2d');
 
-  context?.translate(canvas.width / 2, canvas.height / 2);
+  context?.translate(getHalf(canvas.width), getHalf(canvas.height));
   context?.rotate(rotationRad);
-  context?.drawImage(image, -image.width / 2, -image.height / 2);
+  context?.drawImage(image, getHalf(-image.width), getHalf(-image.height));
```
Some14u:
```suggestion
  context?.drawImage(image, -getHalf(image.width), -getHalf(image.height));
```

Some14u (changes requested):

frontend/src/libs/components/form/libs/helpers/handle-server-error.helper.ts
```diff
@@ -8,7 +8,7 @@ import { ServerErrorType } from '~/libs/enums/enums.js';
 import { type HttpError } from '~/libs/packages/http/http.js';
 import { type FormField } from '~/libs/types/types.js';
 
-import { FIELD_PATH_DELIMITER } from '../consts/consts.js';
+import { PATH_DELIMITER } from '../consts/consts.js';
```
h0wter: I'm guessing it's supposed to be constants?
```suggestion
import { PATH_DELIMITER } from '../constants/constants.js';
```

h0wter (approved):

Some14u (approved):

frontend/src/libs/components/star-rating/star-rating.tsx
```diff
@@ -26,13 +32,13 @@ const StarRating: React.FC<Properties> = ({ rating }: Properties) => {
...
-        stars.push(createStarItem(100, index));
+        stars.push(createStarItem(FULL, index));
       } else if (allStarts >= index) {
-        stars.push(createStarItem((rating - fullFieldStars) * 100, index));
+        stars.push(createStarItem((rating - fullFieldStars) * FULL, index));
       } else {
-        stars.push(createStarItem(0, index));
+        stars.push(createStarItem(EMPTY, index));
       }
     }
```
kalinkaaaa14: is it actually used somewhere? if not, delete it please, because we are not having star rating now

frontend/src/libs/components/file-input/file-input.tsx
kalinkaaaa14: I suppose this comment was more general from those who checked our project, for zero value - ok, I agree that is not the right way 
the most appropriate way for check I think will be
const isNotEmptyRejectedFiles = Boolean(rejectedFiles.length);
 const isEmptyAcceptedFiles = !acceptedFiles.length;

can we do such way @AlexVlaso  ?

kalinkaaaa14 (approved):

PR #370, opened 2023-09-26
th-323: Add error message to the form when truck already exists

denred:


frontend/src/pages/dashboard/components/tables/trucks-table/trucks-table.tsx
```diff
@@ -94,15 +97,19 @@ const TruckTable: React.FC = () => {
           pageSize={pageSize}
           setSorting={setSorting}
           sorting={sorting}
-          isLoading={dataStatus === DataStatus.PENDING}
+          isLoading={dataStatus === DataStatus.IDLE}
```
h0wter: Why is there an IDLE here? In this case the loader will be only once when the component is initialized.

denred: Yes, you're right, this is a mock. I don't think isLoading is needed here at all because the table reloads with each form submission.

h0wter (approved):

AlexVlaso (approved):

PR #387, opened 2023-09-27
th-377:  Send truck geolocation

albsbz:
Add request for driver geolocation and update geolocation on shift start
Finish geolocation update when shift closed
Restart geolocation update when driver page reloaded and shift open


kalinkaaaa14 (approved):

frontend/src/libs/packages/middleware/libs/enums/geolocation-messages.enum.ts
```diff
@@ -0,0 +1,5 @@
+const GeolocationMessages = {
```
mV22110702, line 1:
```suggestion
const GeolocationMessage = {
```

frontend/src/slices/trucks/actions.ts
```diff
@@ -55,6 +55,24 @@ const updateTruckLocationFromSocket = createAsyncThunk<
...
+  },
+);
+
+const restartWatchTruckLocation = createAction(
+  ActionName.RESTART_WATCH_TRUCK_LOCATION,
+  () => {
+    return {
+      payload: null,
+    };
+  },
```
mV22110702, line 73: An action can be created without a payload, so here it is redundant.

frontend/src/libs/packages/middleware/geolocation.middleware.ts
```diff
@@ -0,0 +1,76 @@
...
+  getState: () => RootReducer;
+}) => {
+  const locationUpdateIntervals: NodeJS.Timer[] = [];
+
+  const setLocationUpdateInterval = (interval: NodeJS.Timer): void => {
+    for (const oldInterval of locationUpdateIntervals) {
+      clearInterval(oldInterval);
+    }
+    locationUpdateIntervals.push(interval);
+  };
```
mV22110702, line 40: Why are we using array here? Maybe it'd better without it?

```suggestion
  const locationUpdateInterval: NodeJS.Timer|null = null;

  const setLocationUpdateInterval = (interval: NodeJS.Timer): void => {
      locationUpdateInterval = interval;
  };
```

frontend/src/libs/packages/middleware/libs/helpers/update-geolocation.helper.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+const handleUpdateError = (error: GeolocationPositionError): void => {
+  getLocationError(error.message);
+};
+
+const updateGeolocation = async (
+  payload: StartWatchTruckLocation,
+  setUpdateLocationInterval: (interval: NodeJS.Timer) => void,
+): Promise<boolean | null> => {
+  if (socketInstance) {
```
mV22110702, line 38: Too big if body. It's better to 
```suggestion
  if (!socketInstance) { return null }
```

mV22110702 (changes requested):

mV22110702 (approved):

PR #389, opened 2023-09-28
th-248: Remove Non-nullable From driver_license_file

Some14u:
Hope it fix the issue completely. 
I managed to deal with the problem of server failing to get the driver with unverified status.

### DIFF
--- a/backend/src/packages/drivers/driver.entity.ts
+++ b/backend/src/packages/drivers/driver.entity.ts
@@ -14,7 +14,7 @@ class DriverEntity implements IEntity {
 
   private driverLicenseNumber: DriverEntityT['driverLicenseNumber'];
 
-  private driverLicenseFileId: DriverEntityT['driverLicenseFileId'] | null;
+  private driverLicenseFileId: DriverEntityT['driverLicenseFileId'];
 
   private userId: DriverEntityT['userId'];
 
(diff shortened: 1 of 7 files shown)


kalinkaaaa14 (dismissed):

backend/src/packages/drivers/driver.entity.ts
```diff
@@ -14,7 +14,7 @@ class DriverEntity implements IEntity {
 
   private driverLicenseNumber: DriverEntityT['driverLicenseNumber'];
 
-  private driverLicenseFileId: DriverEntityT['driverLicenseFileId'];
+  private driverLicenseFileId: DriverEntityT['driverLicenseFileId'] | null;
```
mV22110702, line 17: We already have added `|null` in driver-entity.type, so here it's redundant

mV22110702 (changes requested):

backend/src/packages/drivers/driver.entity.ts
mV22110702, line 17: We already have added `|null` in driver-entity.type, so here it's redundant

mV22110702: We already have added `|null` in driver-entity.type, so here it's redundant

mV22110702 (approved):

PavloSukhinin (approved):

mV22110702 (approved):

PR #403, opened 2023-09-28
th-392: Bug when a logged-in customer made an order, his ID is not stored in the database

denred:


frontend/src/packages/orders/orders-api.ts
```diff
@@ -50,7 +50,7 @@ class OrdersApi extends HttpApi {
         method: 'POST',
         contentType: ContentType.JSON,
         payload: JSON.stringify(payload),
-        hasAuth: false,
+        hasAuth: true,
       },
```
albsbz, line 54: not registered user will not be able create order?

denred: This parameter adds only a header to the request. And if not, then how else will the order know that a registered user wants to place it?

denred: Of course, I tested it before the merge.

Druxight (approved):

PR #443, opened 2023-09-30
th-428:  The driver with an active shift does not receive orders

albsbz:
The driver with an active shift does not receive orders


frontend/src/libs/packages/middleware/socket.middleware.ts
```diff
@@ -46,6 +47,7 @@ const socketMiddleware: ThunkMiddleware<
       ServerToClientEvent.ORDER_CREATED,
       (order: OrderResponseDto) => {
         void dispatch(createOrderFromSocket(order));
+        window.location.assign(`/driver/orders/${order.id}`);
```
kalinkaaaa14, line 50: it will be better to use navigate

frontend/src/libs/packages/middleware/socket.middleware.ts
```diff
@@ -87,6 +89,16 @@ const socketMiddleware: ThunkMiddleware<
           break;
         }
 
+        case subscribeDriverOrderCreated.type: {
+          socketInstance.emit(
+            ClientToServerEvent.SUBSCRIBE_DRIVER_ORDER_CREATED,
+            {
+              driverId: `${action.payload as string}`,
```
kalinkaaaa14, line 96: it's unnecessary to cast it to string and use backticks

kalinkaaaa14 (approved):

